PR #1, opened 2022-12-13
v0.1.0 improvements

dimitropoulos:
and other associated packaging updates (including aligning more with tanstack)


.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
+  "singleQuote": true,
```
paul-sachs, line 2: Curious why we chose to use singleQuote instead of our usual in our other repos. Not that we have a specific style across our current repos but the connect repos default to the default prettier config.

packages/connect-query/src/unary-hooks.test.ts
```diff
@@ -741,15 +743,15 @@ describe('unaryHooks', () => {
         >;
       });
 
-      it('passes through getPlaceholderData, when provided', () => {
+      it('passes through placeholderData, when provided', () => {
```
paul-sachs, line 746: Super nit: It's not really passing through placeholderData as much as it is running the right `getPlaceholderData` to generate the right placeholderData

.prettierrc
dimitropoulos: https://github.com/TanStack/query/blob/4abce31572171b9b33886f4680adcc780eeec45c/.prettierrc#L3

paul-sachs: We also discussed this offline. In keeping with the style of tanstack, this makes sense.

packages/connect-query/src/unary-hooks.test.ts
dimitropoulos: much appreciated! 0ac91b66868f00828712439ca13273d69e3968bb

dimitropoulos:
closing as to rename the branch

PR #15, opened 2023-01-19
Updated slack links

smaye81:
This updates all the links to Slack to our main link which is maintained to always redirect to an active link.


dimitropoulos (approved):
thanks! I got this from the contact page originally, and didn't know it expires.

PR #19, opened 2023-02-13
handles JavaScript reserved words

### LINKED ISSUE #18: `const ${localName(method)}` can result in a not allowed variable name
I have a case where a service has a "Delete" method. Since the generator will output `export const ${localName(method)} = ...`, it will output `export const delete = ...`, which is a reserved keyword in javascript and cannot be used as a variable name in strict mode. This will make typescript compilation fail with no way of ignoring the issue.

While "delete" is the more likely case to happen in this context, all methods named after a [reserved word](https://www.w3schools.com/js/js_reserved.asp) will also generate invalid code as such.

A simple solution could be using PascalCase for the variable declaration, but that would also break any code using this after the upgrade.

What are your thought on this? 

In the meantime i'll just automatically patch the file after generation with a sed command.

Thanks a lot by the way, this toolchain is a game changer for us compared to using grpc-web! :)

dimitropoulos:
closes #18 

It seems like our options are:
1. throw when trying to serialize identifiers that will result in syntax errors
2. do something to the identifiers so that they're not syntax errors
  - this will, of course, introduce inconsistencies for special cases where these reserved words have been used
  -  I am pretty open minded on what the specific suffix should be.


dimitropoulos:
closing in favor of https://github.com/bufbuild/connect-query/pull/28

PR #22, opened 2023-02-13
Updates all dependencies

dimitropoulos:
to make it repeatable and easy for the future, I've added a simple root package script, `update-all`, (and documented its usage in a `CONTRIBUTING.md`).


packages/generated-react/package.json
```diff
@@ -23,11 +23,11 @@
     "generate:descriptorset": "buf build --path eliza.proto --path example.proto --output dist/descriptorset.bin"
   },
   "dependencies": {
+    "@bufbuild/connect-query": "workspace:^0.1.0",
     "@bufbuild/protobuf": "^1.0.0",
-    "@bufbuild/protoc-gen-connect-query": "*",
-    "@bufbuild/protoc-gen-connect-web": "^0.6.0",
-    "@bufbuild/protoc-gen-es": "^1.0.0",
-    "@bufbuild/connect-query": "*"
+    "@bufbuild/protoc-gen-connect-query": "workspace:^0.1.0",
```
dimitropoulos, line 28: in retrospect, this is better in pretty much every way I can think of and I probably should have done this in the first place.  good on `pnpm` for making the improvement automatically!

paul-sachs: Oh, that was automatic? Very cool. Yeah workspace is a much better prefix to avoid installing a malicious package.

paul-sachs (approved):

PR #26, opened 2023-03-03
Updates all dependencies

dimitropoulos:
This one was tricky due to some breaking changes in connect-es.  While I'm not terribly happy with the solution, it appears to work, the required changes for the test `useMutation -> makes a mutation` in `unary-hooks.ts` bothers me a lot.


packages/connect-query/jest.config.ts
```diff
@@ -16,7 +16,8 @@ import type { Config } from 'jest';
 
 const config: Config = {
   preset: '../../jest-preset.js',
-  testEnvironment: 'jsdom',
+  // testEnvironment: 'jsdom',
```
paul-sachs, line 19: Maybe delete comment... Or keep it, up to you.

packages/connect-query/src/unary-hooks.test.ts
```diff
@@ -878,8 +878,8 @@ describe('unaryHooks', () => {
         expect(result.current.get.isSuccess).toBeTruthy();
       });
 
-      expect(result.current.get.data?.count).toStrictEqual(1n);
       expect(result.current.mut.data).toStrictEqual(undefined);
+      expect(result.current.get.data?.count).toStrictEqual(0n);
```
paul-sachs, line 882: Did these tests need to change? Not sure i understand why they would...

paul-sachs (approved):
Very minor questions, nothing blocking.

packages/connect-query/src/unary-hooks.test.ts
dimitropoulos: SAME.  that's why I called it out.  This is what it takes to make them pass and I'm pretty confused as to why.

packages/connect-query/jest.config.ts
dimitropoulos: good call, I missed this.

packages/connect-query/src/unary-hooks.test.ts
dimitropoulos: update: paul and I talked and decided to move forward.  we have other things than this to investigate going forward next week, anyway.

PR #28, opened 2023-03-28
updates all dependencies

dimitropoulos:
now that https://github.com/bufbuild/protobuf-es/pull/391 has been released in https://github.com/bufbuild/protobuf-es/releases/tag/v1.2.0, we can use it here to address exports which happen to collide with JavaScript reserved words.

closes https://github.com/bufbuild/connect-query/issues/18


examples/react/basic/tsconfig.json
```diff
@@ -1,20 +1,21 @@
...
-    "target": "ESNext",
-    "useDefineForClassFields": true,
-    "lib": ["DOM", "DOM.Iterable", "ESNext"],
     "allowJs": false,
-    "skipLibCheck": true,
-    "esModuleInterop": false,
     "allowSyntheticDefaultImports": true,
-    "strict": true,
+    "esModuleInterop": false,
+    "exactOptionalPropertyTypes": true,
```
timostamm, line 6: Are we making some useful changes to the example?

tsconfig.base.json
```diff
@@ -2,7 +2,6 @@
   "compilerOptions": {
     "target": "ESNext",
     "esModuleInterop": true,
-    "importsNotUsedAsValues": "error",
```
timostamm, line 5: Relying on eslint instead, or is there a different reason for the change?

timostamm (commented):
Looks good to me. Just a question about the TSC changes.

tsconfig.base.json
dimitropoulos: yep: it's deprecated now in TS 5

timostamm: 👍 

Just catching up on the release notes...  [verbatimModuleSyntax](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#verbatimmodulesyntax) is nice for friends of ESM. This will ruffle some feathers in various bundlers.

examples/react/basic/tsconfig.json
timostamm: I was literally wondering if this is a change or just a re-ordering. Looks like we're enabling exactOptionalPropertyTypes.

dimitropoulos: ahh ok: it's not functionally (sorting aside) a change because the codebase already follows exactOptionalProperties, I just hadn't realized that this wasn't on for this example (I tried to keep the examples simpler and more portable by not inheriting from the base).

PR #34, opened 2023-04-04
Move all tests to `createRouterTransport` from connect

dimitropoulos:
The prior approach this package took to testing was a means to an end, but came with some issues.

We here at Buf put our heads together and came up with `createRouterTransport` which allows testing simulated connect backends without having to run a webserver, but still going though the same codepaths the code _would_ go through in production.

This is a vast improvement.  (more documentation on "how to do testing with Connect Query to come in the future!)


dimitropoulos:
@paul-sachs and that's not even the best part! lol.  it also WORKS much better and is actually _correct_.  As you may have noticed, it turned out with the "mocking fetch" approach we were returning things in one or two small cases that a real Connect server wouldn't _quite_ return.  That, to me, is all the justification needed!  If you can't trust your tests... haha.. what _can_ you trust :small_airplane:?

srikrsna-buf (approved):

PR #35, opened 2023-04-04
Updates useQuery and createUseQueryOptions to reflect undefined returned from queryFn

dimitropoulos:
closes https://github.com/bufbuild/connect-query/issues/29

(I have written some tests but they're not included here because they'll need to come after https://github.com/bufbuild/connect-query/pull/34)


dimitropoulos:
update: actually it's already working as expected, the issue in question ended up being about `strictNullChecks` needing to be on.

PR #20, opened 2023-02-13
Ship esm for connect-query (in addition to cjs)

dimitropoulos:
closes #17 

It doesn't seem like this needs to also be done for the protoc generator, but please someone correct me if I'm wrong.

PR #36, opened 2023-04-05
chore: migrate build tooling to unbuild

getinnocuous:
In relation to #20

Using unbuild to generate both ESM and CJS files


dimitropoulos:
dang, I wish I had come up for air sooner and seen this PR.  We were able to get things working with tsup, but the downside is that we're still shipping CJS, which is non-ideal.

So, I updated your PR and tried to get it working.  I noticed that if I `pnpm pack` and pass it to arethetypeswrong it gives me this:



any ideas?  I'd be happy to switch to unbuild if it gives us good/correct output for both so that we can ship native esm.

getinnocuous:
> dang, I wish I had come up for air sooner and seen this PR. We were able to get things working with tsup, but the downside is that we're still shipping CJS, which is non-ideal.
> 
> So, I updated your PR and tried to get it working. I noticed that if I `pnpm pack` and pass it to arethetypeswrong it gives me this:
> 
> 
> 
> any ideas? I'd be happy to switch to unbuild if it gives us good/correct output for both so that we can ship native esm.

I tried using the [unjs template package.json](https://github.com/unjs/template/blob/main/package.json#L9) and this gives the following output. CJS still looks like an issue but my knowledge of CJS vs ESM is fairly limited so I can't really offer any advice on what the best approach from here would be.

Screenshot 2023-04-28 at 11 21 22

```
"type": "module",
"exports": {
  ".": {
    "types": "./dist/index.d.ts",
    "import": "./dist/index.mjs",
    "require": "./dist/index.cjs"
  }
},
"main": "./dist/index.cjs",
"module": "./dist/index.mjs",
"types": "./dist/index.d.ts",
```

dimitropoulos:
ok cool cool.  no worries then :)

I'll keep my eyes open in the future if I see an opportunity to do this.  thanks for the PR!  I'll probably revisit in the future if we get interested in shipping native ESM.

PR #41, opened 2023-05-09
Docs for streaming

dimitropoulos:
answers https://github.com/bufbuild/connect-query/issues/31

This PR aims to answer the question of "what's the current status of streaming with Connect-Query".


paul-sachs (approved):
Noice! It does make sense to document this current limitation. I thought I had seen some docs in TanStack at some point about streaming but I can't find them. Might have been imagining.

PR #47, opened 2023-05-24
Allow fallback of pageParamKey to current input

paul-sachs:
Ensure that the pageParam is not overriding the pageParamKey on initial request.

I also added a paginated service to more easily test paginated APIs.


paul-sachs:
This PR is related to #47 and should address the immediate need to prevent us from overriding the page key. A follow up PR will introduce a more flexible approach to applying the pageParam.

jest-preset.js
```diff
@@ -30,6 +30,7 @@ module.exports = {
   snapshotFormat: {
     printBasicPrototype: false,
   },
+  workerThreads: true,
```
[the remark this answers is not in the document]
dimitropoulos: all the tests passed for me even when turning this off -> maybe it isn't needed anymore?  I've seen this problem before too but usually it was when I was doing `expect(1).toBe(1n)` or something.  When the types match, it seems to work fine.  Take another look please, but if not please add your comment in the code so future readers will know why it's needed.

```
  workerThreads: true, // Enabled so jest would stop complaining about serializing BigInt.  See https://github.com/jestjs/jest/issues/11617#issuecomment-1458155552 for details
```

packages/connect-query/src/jest/test-utils.tsx
```diff
@@ -139,6 +140,24 @@ export const mockStatefulBigIntTransport = () =>
...
+ * a mock for PaginatedService that acts as an impromptu database
+ */
+export const mockStatefulPaginatedTransport = () =>
+  createRouterTransport(({ service }) => {
+    let base = 1n;
+    service(PaginatedService, {
+      list: (request) => {
+        const result = {
+          page: request.page,
+          items: [`${base} Item`, `${base + 1n} Item`, `${base + 2n} Item`],
```
dimitropoulos, line 153: I love this.  reallllly elegant and simple solution to test what we're trying to test!

jest-preset.js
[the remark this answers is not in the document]
paul-sachs: It only passes because there are no errors in the tests. If you disable this flag and then cause a comparison that fails with a bigint, it'll fail again. Adding a comment makes sense since it is far from obvious.

packages/connect-query/src/jest/test-utils.tsx
paul-sachs: Hmm, actually just noticed a bug in it. We should have base be the same as the requested page. That way a subsequent request for page 1 after a request for page 2 will actually return the items.

packages/connect-query/src/unary-hooks.ts
```diff
@@ -276,6 +277,8 @@ export const unaryHooks = <I extends Message<I>, O extends Message<O>>({
             input !== disableQuery,
             'queryFn does not accept a disabled query',
           );
+          const valueAtPageParam =
+            input[pageParamKey as unknown as keyof PartialMessage<I>];
```
dimitropoulos: this need to cast suggests to me that the original type was just wrong.  after all: does it really matter if we do ParitalMessage instead of PlainMessage for pageParamKey itself?  Maybe this is more correct this way 0fc87fb6110680afa55b5b36990adb75b75dd0da.  Take a look (the intention is that if we don't like it for any reason we can drop the commit).

packages/generated-react/dist/example-TodoService_connectquery.ts
```diff
@@ -50,7 +50,7 @@ export const addTodo = createQueryService({
  *
  * @generated from rpc buf.connect.demo.example.v1.TodoService.Delete
  */
-export const delete$ = createQueryService({
+export const delete = createQueryService({
```
dimitropoulos, line 53: uh oh.  what's going on here.  this is invalid syntax in javascript.  when I did a `pnpm all` it overwrote these, fixing this as well as the bad import at the top that's missing `@bufbuild/`.  I'd LOVE to understand how this happened.  so strange.

dimitropoulos: I see above that it (somehow?) ran with 0.2.1 instead of 0.2.2.  not sure if/how that's related.

packages/connect-query/src/unary-hooks.ts
paul-sachs: Oh man, no that looks great. I struggled a little bit with that and just wanted to make sure it works before returning to it. Makes total sense.

packages/connect-query/src/jest/test-utils.tsx
dimitropoulos: haha, well I stand corrected then!  but either way I like the concept!

packages/connect-query/src/unary-hooks.ts
dimitropoulos: oh cool, sweet.  I wasn't sure but this also seems good to me too.  let's keep the commit then until someone tells us otherwise (in the future, I mean, beyond this PR).

jest-preset.js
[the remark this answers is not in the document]
dimitropoulos:
```suggestion
  workerThreads: true, // Enabled so jest would stop complaining about serializing BigInt.  See https://github.com/jestjs/jest/issues/11617#issuecomment-1458155552 for details
```

packages/connect-query/src/jest/test-utils.tsx
paul-sachs: Alright, modified. Now it's properly non-stateful.

packages/generated-react/dist/example-TodoService_connectquery.ts
paul-sachs: Hmm, weird. Yeeeah. All i did was run `pnpm generate` and that's what popped out.

paul-sachs: Just ran `pnpm all` and it cleaned it up... huh..

paul-sachs: I did run `pnpm install` before this but maybe something borked up along the way.

PR #50, opened 2023-05-30
[experiment] attempts to use jest globals instead of a custom environment

dimitropoulos:
(this is an experimental branch to explore the guidance in https://github.com/bufbuild/connect-es/issues/658)


dimitropoulos:
the above passing CI is very suspicious because [it doesn't pass locally](https://github.com/bufbuild/connect-query/actions/runs/5121825226/jobs/9210074649?pr=50).

Screenshot 2023-05-30 at 09 09 33

timostamm:
The DataCloneError seems to confirm the suspicion in https://github.com/bufbuild/connect-es/issues/658#issuecomment-1576410472, that Jest clones globals to prevent individual tests from modifying it.

Thanks for confirming that globals won't work 👍  We will look into publishing an environment as an npm package.

PR #52, opened 2023-06-12
Cleanup dependencies

paul-sachs:
- Cleanup deps
- Remove unnecessary deps and move protobuf to peer

### DIFF
--- a/packages/connect-query/package.json
+++ b/packages/connect-query/package.json
@@ -38,7 +38,8 @@
     "@tanstack/react-query": "^4.20.4",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
-    "@bufbuild/protobuf": ">=1.2.0"
+    "@bufbuild/protobuf": ">=1.2.0",
+    "@bufbuild/connect": ">=0.9.1"
   },
   "dependencies": {
     "stable-hash": "^0.0.3"
--- a/packages/connect-query/package.json
+++ b/packages/connect-query/package.json
@@ -29,7 +29,10 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "tsup": "^6.7.0",
-    "typescript": "^5.0.4"
+    "typescript": "^5.0.4",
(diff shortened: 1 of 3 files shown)


packages/connect-query/package.json
```diff
@@ -29,18 +29,18 @@
...
     "react": "^18.2.0",
-    "react-dom": "^18.2.0"
+    "react-dom": "^18.2.0",
+    "@bufbuild/protobuf": ">=1.2.0"
   },
   "dependencies": {
-    "@bufbuild/connect": "^0.9.0",
-    "@bufbuild/connect-web": "^0.9.0",
-    "@bufbuild/protobuf": "^1.2.0",
-    "@bufbuild/protoc-gen-connect-es": "^0.9.0",
```
timostamm, line 43: From a look at the source of @bufbuild/connect-query, we are importing from @bufbuild/protobuf, but also from @bufbuild/connect. For example, see here: https://github.com/bufbuild/connect-query/blob/5553f604af2b3135816ab54a5296ff86cbec779e/packages/connect-query/src/use-transport.tsx#L15-L16

That means we _have_ to declare a dependency on @bufbuild/connect, doesn't it? What's the appropriate dependency here - direct, or peer? 

I don't think we are importing from @bufbuild/connect-web - that's on the user when setting up the transport ([example](https://github.com/bufbuild/connect-query/blob/5553f604af2b3135816ab54a5296ff86cbec779e/examples/react/basic/src/main.tsx#LL18C10-L18C32)).

paul-sachs: Ah I missed that... Man I wish it was easier to write tests for these kinds of things. I would prefer if we add it as a peer, making sure that any `instanceof ConnectError` would return true appropriately. The problem is really the install gets more complicated because of it...

paul-sachs: Or so I had assumed, I guess npm@7+ auto installs peerDependencies, though still not my favourite.

timostamm (approved):

PR #59, opened 2023-07-12
Feat: Additional plugin to generate connect-query with hooks

vctqs1:
Wrapping react-query directly. By default useQuery, useMutation, useInfiniteQuery will be import from @tanstack/react-query til we provide opts
```
- opt: import-hook-from=@custom/react-query
```

Generate code sample
```


export const useDelete$Query = 
  (...inputs: Parameters<typeof delete$.useQuery>) => useQuery(delete$.useQuery(inputs));

export const useDelete$Mutation = 
  (...inputs: Parameters<typeof delete$.useMutation>) => useMutation(delete$.useMutation(inputs));

export const useDelete$InfiniteQuery = 
  (...inputs: Parameters<typeof delete$.useInfiniteQuery>) => useInfiniteQuery(delete$.useInfiniteQuery(inputs));

```


paul-sachs:
Thanks @vctqs1 for contributing!

Pretty cool idea to make `connect-query` a little less verbose. connect-query was created as a base building block to allow exactly this kind of thing. I like the idea but one of the things I often see is also modifying the params sent to apis like `useMutation`. For example, I often want toasts:

```tsx
const showToast = useToast();
const mutation = useMutation({
  ...someMutation.useMutation(),
  onSuccess: () => {
    showToast("Success!");
  }
});
```

The API you've defined above would not allow this. If we expand the parameters to also support these options, then I think it would create a very robust API. Perhaps something that created an API like:

```ts
export const useCountQuery = 
  ({ inputs, transformParams }:
    { 
      inputs: Parameters<typeof count.useQuery>, 
      transformParams?: (params: ReturnType<count.useQuery>) => QueryOptions<RequestMessage, Error, ...>
    }) => {
      const baseOptions = count.useQuery(inputs)
      return useQuery(transformParams ? transformParams(baseOptions) : baseOptions);
    }
```

Haven't typed this fully yet, would need to figure out `QueryOptions<RequestMessage, Error, ...>`. Which makes it easy expand the options sent to `useQuery`.

The above is definitely not the only possible API, just thinking through some options.

packages/generated-react/dist/eliza-BigIntService_connectquery_hooks.ts
```diff
@@ -0,0 +1,33 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// @generated by protoc-gen-connect-query-with-hooks v0.0.1 with parameter "target=ts,import-hook-from=@tanstack/react-query"
+// @generated from file eliza.proto (package buf.connect.demo.eliza.v1, syntax proto3)
+/* eslint-disable */
+// @ts-nocheck
+
+import { count } from "./eliza-BigIntService_connectquery.ts";
```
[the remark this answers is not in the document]
vctqs1: > This import means that this plugin cannot work without the connect-query plugin also running.

I think this is trade-off. Because this plugin need to call `connect-query` under hood somehow, i think it's make our plugins be in a line together.

> since if someone specifies ONLY the hook version, imports will fail.
Yeah, that why I was put it in README to install both connect-query and hooks plugin

> I think it would be a huge advantage to not have to rely on that
@paul-sachs could you help suggest on this. I've no clue how to make it work without connect-query/ please help me :+1:

vctqs1:
> ```ts
>       transformParams?: (params: ReturnType<count.useQuery>) => QueryOptions<RequestMessage, Error, ...>
> ```

I'll address it now. Thank you. This is nice suggestion

packages/generated-react/dist/eliza-BigIntService_connectquery_hooks.ts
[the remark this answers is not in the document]
timostamm: @vctqs1, what Paul means is that the code generated by protoc-gen-connect-query-with-hooks requires three other plugins to work:

```yaml
plugins:
  - name: es
    out: dist
    opt:
      - target=ts
      - import_extension=.js
  - name: connect-web
    out: dist
    opt:
      - target=ts
      - import_extension=.js
  - name: connect-query
    opt: target=ts
    out: dist
  - name: connect-query-with-hooks
    opt: 
      - target=ts
      - import-hook-from=@tanstack/react-query
    out: dist
```

It would be a better user experience if the new plugin did not rely on code generated by protoc-gen-connect-query, so that you would simply use protoc-gen-connect-query if you want the standard behavior, protoc-gen-connect-query-with-hooks if you want the hooks, but not both at the same time.

The best way to do that is do duplicate code in the plugin where necessary. Instead of importing `count`, generate `count`.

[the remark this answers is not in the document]
paul-sachs: Yeah, you'd basically want to copy the content from [here](https://github.com/bufbuild/connect-query/blob/main/packages/protoc-gen-connect-query/src/generateTs.ts) and do roughly the same thing. You can probably move the jsDoc and export since that's not necessary, but you'll basically want to call `createQueryService` from `@bufbuild/connect-query` in your generated code.

packages/generated-react/dist/eliza-BigIntService_connectquery_hooks.ts
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * @generated from rpc buf.connect.demo.eliza.v1.BigIntService.Count
+ */
+export const useCountQuery = 
+  ({
+    inputs,
+    transformParams,
+  }: {
+    inputs: Parameters<typeof count.useQuery>;
```
paul-sachs, line 34: I was probably a little hasty in building this API, given that this is what it would look like in usage:

```tsx
const q = useCountQuery({
  inputs: [{ count: 1 }],
});
```
for the common case. I think maybe we can do better for both common cases, and when you want to modify the params.

`useQuery` from `connect-query` always has two arguments, one for the input and the second is options. So maybe we do:

```tsx
export const useCountQuery = 
  (input: Parameters<typeof count.useQuery>[0], options: Parameters<typeof count.useQuery>[1], queryOptions?: {
    transformParams?: (
      baseOptions: ReturnType<typeof count.useQuery>
    ) => Partial<UseBaseQueryOptions<PartialMessage<CountRequest>, ConnectError>>;
  }) => {
    const { transformParams } = queryOptions ?? {};
    const baseOptions = count.useQuery(input, options);
```

The above is a little more verbose, but it allows a 1:1 swap of apis:

```tsx
const q = useCountQuery({ count: 1 });
const q2 = useCountQuery({ count: 1}, { transport, transformParams });
```

We could merge queryOptions and options but that creates potential for confusion if we ever introduce an option called `transformParams` to `useQuery` options so this is safest.

packages/generated-react/dist/eliza-BigIntService_connectquery_hooks.ts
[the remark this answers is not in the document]
paul-sachs: @timostamm I'd want to build ontop of this base to add those APIs. There are more streamlined apis that don't require `useQueryClient` calls all over the place. There are APIs like `invalidateQueries` which would be really helpful that `UnaryHooks` doesn't implement directly that could be done here. But an incremental approach makes sense to me.

packages/generated-react/dist/eliza-BigIntService_connectquery_hooks.ts
paul-sachs: Though honestly, merging the options objects should be fine. I think the chance for overlap is very slim.

packages/generated-react/dist/eliza-BigIntService_connectquery_hooks.ts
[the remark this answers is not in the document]
vctqs1: Thank you, I got you two idea.

packages/generated-react/dist/eliza-BigIntService_connectquery_hooks.ts
vctqs1: ah yeah, using 
```
inputs: [
            {
               ..variables,
            },
            {},
        ],
```
Kinda annoy and a bit ugly. Let me do adjust

vctqs1: @paul-sachs for `queryOptions` is it okay to pass exactly type from `react-query` instead our custom type `{ transformParams }`

EX i'll look like
```
useCountQuery({count: 1}, {transport}, {onError, onSuccess})
```

paul-sachs:
@vctqs1 I was thinking this through and I think it should rename the plugin to `protoc-gen-connect-query-react` since it's basically creating opinionated hooks are are specific to react. That way we can start adding more react specific helpers to this as well. What do you think?

vctqs1:
> @vctqs1 I was thinking this through and I think it should rename the plugin to `protoc-gen-connect-query-react` since it's basically creating opinionated hooks that are specific to react. That way we can start adding more react specific helpers to this as well. What do you think?

It is nice, I think it makes sense. i got your point. Let me change the name.

packages/generated-react/dist/eliza-BigIntService_connectquery_hooks.ts
vctqs1: yeah. I got your idea. I predicted `queryOptions` be use more than `options`. So place it to second parameter is better, I think

vctqs1: @paul-sachs I updated. could u help review

paul-sachs:
@vctqs1 Alright, I think i've resolved the issues with pnpm lock file in main so we should be good.

vctqs1:
> @vctqs1 Alright, I think i've resolved the issues with pnpm lock file in main so we should be good.

Thank you, please help me know if i've anything need to be resolved in this PR

paul-sachs (approved):
@vctqs1 everything looks good! We'll need to tweak the readme a little because we want to advertise this as an experiment to provide better DX over the existing plugin but we'll handle that after it's merged.

Really appreciate you submitting this PR.

PR #60, opened 2023-07-14
Use corepack to guarantee pnpm version

paul-sachs:
Using corepack should ensure that the correct versions are used.


srikrsna-buf (approved):

srikrsna-buf (commented):
I think the GH action already uses the correct version

paul-sachs:
It does, but this makes sure the GH action will match local runs as well.

PR #64, opened 2023-07-19
Update readme with some references to react specific plugin

paul-sachs:
Calling out the new `protoc-gen-connect-query-react` plugin.

Also introducing `d.ts` version of the plugin and fixes some naming issues.


README.md
```diff
@@ -414,6 +415,19 @@ For example, a partial query key might look like this:
 ]
 ```
 
+# Experimental plugin
+
+There is an alternate plugin (`@bufbuild/protoc-gen-connect-query-react`) that you can use if you are using `@tanstack/react-query` only (and is not compatible with other frameworks like Solid). This plugin is currently experimental to see if it provides a better developer experience.
+
+```tsx
+import { useExampleQuery } from 'your-generated-code/example-ExampleService_connectquery';
```
vctqs1, line 423:
```suggestion
import { useExampleQuery } from 'your-generated-code/example-ExampleService_connectquery_react';
```

paul-sachs: Ah good catch.

README.md
```diff
@@ -414,6 +415,19 @@ For example, a partial query key might look like this:
 ]
 ```
 
+# Experimental plugin
+
+There is an alternate plugin (`@bufbuild/protoc-gen-connect-query-react`) that you can use if you are using `@tanstack/react-query` only (and is not compatible with other frameworks like Solid). This plugin is currently experimental to see if it provides a better developer experience.
```
timostamm, line 420: We have to link to the detailed documentation here. I'd link to https://www.npmjs.com/package/@bufbuild/protoc-gen-connect-query-react

packages/protoc-gen-connect-query-react/README.md
```diff
@@ -1,5 +1,7 @@
 # @bufbuild/protoc-gen-connect-query-react
 
+**This is an experimental plugin, based on the [protoc-gen-connect-query](../protoc-gen-connect-query/README.md) plugin. This plugin provides better DX while sacrificing some flexibility.**
```
timostamm, line 3: Can we point out the main difference to protoc-gen-connect-query? It generates hooks for react.

We can't link to `../protoc-gen-connect-query/README.md` here because that doesn't resolve on npmjs.com. Let's link to https://www.npmjs.com/package/@bufbuild/protoc-gen-connect-query instead.

timostamm (commented):
Before we publish, please make two changes to the [plugin readme](https://github.com/bufbuild/connect-query/blob/aaad89b7f8a12892c8a9d0fbe8dbd85b9e96aa0c/packages/protoc-gen-connect-query-react/README.md):

1. All file examples have the `_connectquery` suffix instead of `_connectquery_react`
2. We never explain what the plugin does - at the very least, we have to show a snippet of code that illustrates that a query hook is generated.

And there is a bug: 

`export const useDelete$Query = ...` (see [here](https://github.com/bufbuild/connect-query/blob/bf1ceb9625c5d247f01d2009e8d80df53721e472/packages/generated-react/dist/example-TodoService_connectquery_react.ts#L143C3-L143C3)) unnecessarily escapes `delete` by appending `$`. This is necessary for `protoc-gen-connect-query` because it would otherwise generate `export const delete = ...`, which is a reserved word and would fail to compile. This should be fixed before we publish because the fix would be a breaking change for users right away.

And the plugin does not have a declaration file generator. That means that it falls back to using tsc to transpile a .d.ts file if requested. That's okay for a project-specific plugin, but for a plugin we publish, we have to make sure that it works for declaration files as well.

packages/protoc-gen-connect-query-react/src/generateTs.ts
```diff
@@ -102,38 +103,12 @@ const generateServiceFile =
           importHookFrom,
         );
 
-        f.print(
```
paul-sachs, line 105: Removed prettier formatting on this since it's much easier to read without.

timostamm: I agree. Prettier is not made for formatting code generators 🙂

timostamm (approved):
Noticed that we still had a `const useValueOf$Query`. Pushed up https://github.com/bufbuild/connect-query/pull/64/commits/7e662f05d3cb1dfb6bad09d1cf5ee5da09cf7b62 that adds a function `reactHookName` to construct the name from the original proto RPC name, so that it's `const useValueOfQuery` now.

TY for the updates and the addition of d.ts!

PR #116, opened 2023-08-16
Update all packages to use the @connectrpc scope

timostamm:
To keep Connect well-maintained and responsive to its users' needs over the long term, we're preparing to donate it to a foundation. (More details on that soon!) To cleanly separate Connect from Buf's other code, we're moving development to the [connectrpc](https://github.com/connectrpc) GitHub organization and to the [connectrpc](https://www.npmjs.com/org/connectrpc) organization on npmjs.com.


paul-sachs (approved):
Tentative approval until we get new packages

PR #122, opened 2023-08-23
Add RELEASING.md

timostamm:
Add a guide for the release process in .github/RELEASING.md. Also moves CONTRIBUTING.md to .github and updates it to use the same as other @connectrpc repositories.


.markdownlint.yaml
```diff
@@ -19,11 +19,6 @@ MD013: false
 # Although this rule is great in spirit, the rendered on markdown for such a large root readme simply (visually) requires multiple `h1`s as to decrease one level of nesting so horizontal dividers are easier to see and headings are larger.
 MD025: false
 
-# MD029 - Ordered list item prefix
-# Plain and simple, markdown parsers do not respect the numbering you use for ordered lists.  It only leads to confusion and misconception to allow otherwise.
-MD029:
-  style: one
-
```
timostamm, line 26: The point of markdown is to be readable as plain text. This change reverts to standard markdownlint rules, which allow us to use sequential numbers in plain text for better readability. The linter enforces correct order.

PR #151, opened 2023-09-07
Fix circular dependencies and migrate to turbo

paul-sachs:
Finally got fed up with the build/test cycle not doing what I wanted. Matching format of connectrpc/connect-playwright

- Remove requirement to manually test plugins
- Fixed circular dependencies and migrated to turbo
- More turbo


turbo.json
```diff
@@ -0,0 +1,22 @@
...
+      "dependsOn": ["^build"],
+      "outputs": ["dist/**"]
+    },
+    "generate": {
+      "dependsOn": ["^build"],
+      "inputs": ["*", "!dist"],
+      "outputs": ["dist/**"]
+    },
+    "test": {
+      "dependsOn": ["generate"],
```
srikrsna-buf, line 15: This should also depend on `build`?

paul-sachs: Doesn't need to, since generate depends on `^build` already.

package.json
```diff
@@ -3,18 +3,16 @@
...
-    "lint:markdown:fix": "markdownlint-cli2-fix \"**/*.md\" \"#**/node_modules\"",
-    "lint:spelling": "cspell \"**\" --gitignore",
-    "lint:code": "eslint . && prettier . --check && license-header --check",
-    "lint:code:fix": "eslint . --fix && prettier . --write && license-header",
-    "all": "pnpm run build && pnpm run generate && pnpm run lint:fix && pnpm run test",
-    "update-all": "pnpm outdated --recursive || pnpm update --recursive --latest && pnpm outdated --recursive"
+    "build": "turbo run build",
+    "generate": "turbo run generate",
+    "test": "turbo run test",
+    "format": "turbo run format:all",
```
timostamm, line 9: Not the end of the world, but `pnpm run format` fails when packages have not been built before. You can see it with 

```
git clean -Xdf
pnpm i
pnpm run format
```

I don't think `"dependsOn": ["^generate"]` works.

In my experience, turbo works best when each package uses it's own scripts. It's extra set-up, for sure.

Did you know you can run turbo in a package? For example, you can run `turbo test` in examples/react/basic, and never have to think about what to build first, and it will never do unnecessary work.

package.json
```diff
@@ -3,18 +3,16 @@
...
+    "format:all": "pnpm run format:code && pnpm run format:markdown && pnpm run check:spelling",
+    "format:markdown": "markdownlint-cli2-fix \"**/*.md\" \"#**/node_modules\"",
+    "check:spelling": "cspell \"**\" --gitignore",
+    "format:code": "eslint . --fix && prettier . --write && license-header",
```
timostamm, line 13: If you use eslint ignores, prettier sometimes shifts them. So I would recommend to always format first, then run eslint.

eslint also requires dependencies in the monorepo to be built. My best idea for that would be to make use of turbo dependencies, but that requires you to lint per package. turbo will parallelize when possible though, so it might end up being faster in many situations.

timostamm (approved):
Nice!  I think we might need to break task into smaller pieces to make better use of turbo deps, see comments below. The PR stands on it's own though, so a follow-up should work just as well IMO.

package.json
paul-sachs: I did not know turbo worked in a single package though, that is very cool.

package.json
paul-sachs: I think my only concern for moving all these formatting/linting commands into the packages is how it handles files at the root level (README.md, jest-preset.js, etc)... I guess i could configure the root to run a separate series of commands to format those...

PR #159, opened 2023-09-08
Cleanup

### LINKED ISSUE #11: What is package/generated-react?
Is this an example? We should make this consistent with connect-web/protobuf-es

paul-sachs:
- Simplify examples and move generation into individual packages
- Standardize prettier config with the rest of the connect-* packages
- Made it so protoc-gen-* plugins export the service type so it can be used directly for testing (instead of requiring connect-es to generate it as well)

Fixes #11 

For reviewers, I've added comments to the bulk of the actual changes (everything else is prettier).


packages/protoc-gen-connect-query-react/src/generateDts.ts
```diff
@@ -28,12 +32,34 @@ const { safeIdentifier } = codegenInfo;
...
   (schema: Schema, protoFile: DescFile) => (service: DescService) => {
+    const { MethodKind: rtMethodKind, MethodIdempotency: rtMethodIdempotency } = schema.runtime;
 
     const f = schema.generateFile(
         `${protoFile.name}-${localName(service)}_connectquery_react.d.ts`,
     );
     f.preamble(protoFile);
 
+    f.print(makeJsDoc(service));
+    f.print("export declare const ", localName(service), ": {");
```
paul-sachs, line 43: Added service definition generation here so we can use that for tests directly without needing to build it or rely on connect-es generation as well.

packages/protoc-gen-connect-query-react/src/generateTs.ts
```diff
@@ -46,6 +46,27 @@ const generateServiceFile =
     const { MethodKind: rtMethodKind, MethodIdempotency: rtMethodIdempotency } =
       schema.runtime;
 
+    f.print(makeJsDoc(service));
+    f.print("export const ", localName(service), " = {");
```
paul-sachs, line 50: Ditto, service definition.

packages/protoc-gen-connect-query/src/generateDts.ts
```diff
@@ -12,26 +12,54 @@
...
+    const { MethodKind: rtMethodKind, MethodIdempotency: rtMethodIdempotency } = schema.runtime;
+    
     const f = schema.generateFile(
       `${protoFile.name}-${localName(service)}_connectquery.d.ts`,
     );
 
     f.preamble(protoFile);
 
+    f.print(makeJsDoc(service));
+    f.print("export declare const ", localName(service), ": {");
```
paul-sachs, line 43: Just a bit of copy pasting. Once we prioritize one plugin, we can just delete all of it. That's a diff PR.

turbo.json
```diff
@@ -7,16 +7,14 @@
...
-      "outputs": ["dist/**"]
+      "dependsOn": ["^build"]
     },
     "test": {
       "dependsOn": ["generate"],
       "cache": false
     },
-    "//#format:all": {
-      "dependsOn": ["^generate"]
+    "format": {
```
paul-sachs, line 16: Moving format to individual scripts.

smaye81 (approved):

examples/react/basic/buf.gen.yaml
```diff
@@ -0,0 +1,16 @@
...
+# For details, see https://docs.buf.build/configuration/v1/buf-gen-yaml
+version: v1
+plugins:
+  - name: es
+    path: ./node_modules/.bin/protoc-gen-es
+    out: src/gen
+    opt:
+      - target=ts
+
+  - name: connect-web
```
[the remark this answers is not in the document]
paul-sachs: Ah yeah, sure, why not.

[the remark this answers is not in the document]
paul-sachs: lol, actually it should be connect-query

examples/react/basic/src/gen/eliza-ElizaService_connectquery.ts
```diff
@@ -0,0 +1,112 @@
...
+ * the MIT Artificial Intelligence Laboratory, ELIZA demonstrates the
+ * superficiality of human-computer communication. DOCTOR simulates a
+ * psychotherapist, and is commonly found as an Easter egg in emacs
+ * distributions.
+ *
+ * @generated from service connectrpc.eliza.v1.ElizaService
+ */
+export const ElizaService = {
+  typeName: "connectrpc.eliza.v1.ElizaService",
+  methods: {
```
timostamm, line 45: This is a pretty big change. I assume the export is for `createRouterTransport`?

paul-sachs: Correct. The plugins already generated this, it just wasn't exported. This just makes it so testing can be more easily made without adding another protoc plugin.

paul-sachs: Yeah that's fair. I figured it would make sense for the v1 release. The overlap with `protoc-gen-connect-es` isn't ideal. It does feel like every one of our plugins generates the same structure, I wonder if it makes sense to export a utility from `@bufbuild/protoplugin` since `ServiceType` is defined within `@bufbuild/protobuf`?

timostamm: At this point, and considering that the services are relatively simple, we're better off duplicating the code in the plugins. I'm just thinking that it may be a bit confusing from the user perspective, but as I said before, fair for overall simplicity.

PR #160, opened 2023-09-11
Replace plugin with React specific plugin (v1 prep)

paul-sachs:
Introduces a breaking change (v1) of this library:

- React centric hooks as the primary focus of the plugin
- The core library `@connectrpc/connect-query` now makes even fewer references to react specific constructs like context.
- Also introduces `use<RPC>InvalidateQueries()` for easy access to invalidate queries.

## TODO

- [ ] Docs updates


paul-sachs:
Ready for review, minus documentation changes.

README.md
```diff
@@ -21,9 +22,8 @@ Connect-Query is an expansion pack for [TanStack Query](https://tanstack.com/que
   - [`UnaryHooks.methodInfo`](#unaryhooksmethodinfo)
   - [`UnaryHooks.setQueryData`](#unaryhookssetquerydata)
   - [`UnaryHooks.setQueriesData`](#unaryhookssetqueriesdata)
-  - [`UnaryHooks.useInfiniteQuery`](#unaryhooksuseinfinitequery)
-  - [`UnaryHooks.useMutation`](#unaryhooksusemutation)
-  - [`UnaryHooks.useQuery`](#unaryhooksusequery)
+  - [`UnaryHooks.createUseInfiniteQueryOptions`](#unaryhooksuseinfinitequery)
```
paul-sachs, line 25: Whoops, don't worry about these changes. All these docs need a rework anyways.

.github/workflows/ci.yaml
```diff
@@ -51,6 +51,7 @@ jobs:
 
       - name: Check changed files
         id: check_changed_files
+        continue-on-error: true
         run: |
           git diff --name-only --exit-code
```
smaye81, line 56: I don't know how much we care about consistency, but connect-playwright does a diff check like [this](https://github.com/connectrpc/connect-playwright-es/blob/main/.github/workflows/ci.yaml#L32-L35). Doesn't matter to me which one we use, but we should probs settle on one.

examples/react/basic/buf.gen.yaml
```diff
@@ -9,8 +9,14 @@ plugins:
...
 
   - name: connect-query
-    path: ./node_modules/.bin/protoc-gen-connect-query
+    path: ./node_modules/.bin/protoc-gen-connect-query-react
     out: src/gen
     opt:
       - target=ts
       - import_extension=
+  - name: connect-query
+    path: ./node_modules/.bin/protoc-gen-connect-query-react
```
smaye81, line 18: Did you mean to duplicate these in this file?

.github/workflows/ci.yaml
paul-sachs: I'm all up for a consistent one. I find this very readable without bash gymnastics but my bash skills are also pretty weak so maybe it's just me.

examples/react/basic/buf.gen.yaml
paul-sachs: Yeah, i want to generate both d.ts files as well as .ts files, as a means of detecting the differences. Though in retrospect, perhaps I can put that elsewhere so it doesn't pollute the example 🤔

.github/workflows/ci.yaml
smaye81: Haha, yeah, agreed. But, I don't think `git diff` alerts on untracked files?

```bash
$ touch checkdifftest                                                                                                                                                                                                                                                                                                                                                             
$ git status

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        checkdifftest

nothing added to commit but untracked files present (use "git add" to track)

$ git status --porcelain                                                                                                                                                                                                                                                                                                                                                          
?? checkdifftest

$ git diff --name-only --exit-code
# no output
```

examples/react/basic/buf.gen.yaml
smaye81: Yeah our previous approach on examples in connect-es, et. al. has been to make the examples as straight-forward and understandable as possible. Just a thought though.

.github/workflows/ci.yaml
paul-sachs: oh yeah, good point

PR #165, opened 2023-09-13
Add snapshot tests for plugins

paul-sachs:
Update CI to check for untracked files (as suggested by @smaye81). First PR part of #160.

Few minor adjustments
- fixed a missing (optional) semi-colon in js generation
- added prettier formatting for root level files


packages/protoc-gen-connect-query/snapshots/buf.gen.yaml
```diff
@@ -0,0 +1,22 @@
+# buf.gen.yaml defines a local generation template.
+# For details, see https://docs.buf.build/configuration/v1/buf-gen-yaml
+version: v1
+plugins:
+  - name: es
+    path: ../node_modules/.bin/protoc-gen-es
```
timostamm, line 6: This line can be removed. 

```suggestion
```

For context, with `plugin: es`, the buf CLI tries to use `protoc-gen-connect-es` on your $PATH. 

If we're calling `buf generate` from an npm script or with `npx` or `pnpm exec`, it's run with a modified $PATH that includes the npm "binaries".

In general, we should use the simple form and avoid the full node_modules path. I don't think we can do it for protoc-gen-connect-query _inside_ of this package, because pnpm does not wire it up (it doesn't depend on itself). 

But we can and IMO should use the simple form in the example. If you think it's too much magic, I'd still prefer we use the simple form, but explain it in a comment.

packages/protoc-gen-connect-query/package.json
```diff
@@ -21,10 +21,12 @@
   "scripts": {
     "clean": "rm -rf dist",
     "build": "pnpm clean && tsc --project tsconfig.build.json",
-    "format": "prettier . --write --ignore-path ./.eslintignore && eslint . --fix && license-header"
+    "format": "prettier . --write --ignore-path ./.eslintignore && eslint . --fix && license-header",
+    "generate": "rm -rf snapshots/gen && cd snapshots && buf generate --path eliza.proto"
```
timostamm, line 25: Since we're using the unmodified eliza.proto, `buf generate buf.build/connectrpc/eliza:8b8b971d6fde4dc8ba5d96f9fda7d53c` means we could remove the copy. We end up with more directories though.

Leaving it up to you.

packages/protoc-gen-connect-query-react/tsconfig.json
```diff
@@ -1,4 +1,4 @@
 {
   "extends": "./tsconfig.build.json",
-  "include": ["src"]
+  "include": ["src", "snapshots"]
```
timostamm, line 3: Good call 👍 

Since we're generating both ts and dts - which one will be compiled though? 🤔

packages/protoc-gen-connect-query-react/.eslintignore
```diff
@@ -1 +1,2 @@
-dist
\ No newline at end of file
+dist
+snapshots/**/*.d.ts
```
timostamm, line 2: Is it necessary? There is an eslint-disable comment in every file.

packages/protoc-gen-connect-query/snapshots/buf.gen.yaml
paul-sachs: No that makes sense, i'll adjust.

packages/protoc-gen-connect-query-react/tsconfig.json
paul-sachs: This is really only useful for eslint so it can compile those files. It's a little weird and more finicky than I'd like.

packages/protoc-gen-connect-query-react/.eslintignore
paul-sachs: Yeah this is a little confusing for me as well. Eslint was complaining about processing any file that the tsconfig file didn't process. It only happens for `d.ts` files so maybe I need to adjust the tsconfig file somehow to process d.ts files specifically... I'll dig in a little more.

paul-sachs: It's weird, it's basically because eslint tries to provide the d.ts files to be compiled and our tsconfig doesn't seem to include these files for... reasons? I've made modifications to skip type checking for d.ts files and instead just lint for everything else. Should be ok.

packages/protoc-gen-connect-query-react/tsconfig.json
paul-sachs: This is actually also just for the IDE + test + lint, whereas `tsconfig.build.json` is used for the build itself.

packages/protoc-gen-connect-query/snapshots/buf.gen.yaml
```diff
@@ -0,0 +1,15 @@
...
+    out: gen
+    opt:
+      - target=ts
+
+  - plugin: connect-query
+    path: ../bin/protoc-gen-connect-query
+    out: gen
+    opt:
+      - target=ts+dts+js
+      - import_extension=
```
timostamm, line 15: Same here, use `none`

packages/protoc-gen-connect-query-react/tsconfig.json
timostamm: Gotcha. What I mean is that there is ambiguity in general when you have x.js, x.ts, and x.d.ts all in the same directory, for _any_ tool that processes the files.

I just checked with `npm exec eslint -- snapshots/gen/connectrpc/eliza/v1 --debug`, and see that eslint _does_ lint each file individually.

It would not have caught https://github.com/connectrpc/connect-query-es/issues/166 though. 

That's fine for now - the "snapshots" will make changes reviewable, which is a big improvement. 

Cut https://github.com/connectrpc/connect-es/issues/824, because we don't do better there either.

paul-sachs:
@timostamm Ah ok, my testing was just flawed (changing the wrong plugin). The linter does catch #166:

```
@connectrpc/protoc-gen-connect-query:format: snapshots/gen/connectrpc/eliza/v1/eliza-ElizaService_connectquery.d.ts 33ms
@connectrpc/protoc-gen-connect-query:format: [error] snapshots/gen/connectrpc/eliza/v1/eliza-ElizaService_connectquery.js: SyntaxError: Missing semicolon. (76:2)
@connectrpc/protoc-gen-connect-query:format: [error]   74 |     },
@connectrpc/protoc-gen-connect-query:format: [error]   75 |   }
@connectrpc/protoc-gen-connect-query:format: [error] > 76 | } as const
@connectrpc/protoc-gen-connect-query:format: [error]      |  ^
@connectrpc/protoc-gen-connect-query:format: [error]   77 |
@connectrpc/protoc-gen-connect-query:format: [error]   78 | /**
@connectrpc/protoc-gen-connect-query:format: [error]   79 |  * Say is a unary RPC. Eliza responds to the prompt with a single sentence.
```

timostamm:
> @timostamm Ah ok, my testing was just flawed (changing the wrong plugin). 

And I ran `pnpm exec eslint -- .` to verify, which doesn't catch it for whatever reason...

timostamm (approved):

PR #169, opened 2023-09-13
Extract React specific dependencies into single file

paul-sachs:
Less reliance on React specific APIs (in the core of connect-query) allows the core API to adapt better to scenarios like Server Components or other frameworks.

This PR keeps the existing plugin API (while adding some new methods) but frees us up to be more selective about the react APIs we NEED for other future plugins.

### Breaking changes
- Removed `UnaryHooks` type and replaced with `UnaryFunctionsWithHooks`
- Updated `createUnaryHooks` to only return the hooks and not other functions
 
### createUnaryFunctions

Introduced `createUnaryFunctions` which can be used with the new `createUnaryHooks` to do the same thing as the old `createUnaryHooks`. The reasoning behind this change is to prep the core library, `connect-query`, to be able to used separately from React itself and this is just the first step in that direction.

### New functions

Each service method now generates two new functions associated to the method:

#### method.createUseInfiniteQueryOptions

This method creates options that will pass to `useInfiniteQuery`. It is identical to `method.useInfiniteQuery` except it needs to be passed a transport directly.

#### createUseMutationOptions

This method creates options that will pass to `useMutation`. It is identical to `method.useMutation` except it needs to be passed a transport directly.

We added these methods to better future-proof the code for incoming changes to React (React Server Components, etc).

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,11 +2,12 @@
 <img src="assets/connect-query@16x.png" width="15%" />
 
 <!-- omit in toc -->
+
 # Connect-Query
 
 [![License](https://img.shields.io/github/license/connectrpc/connect-query-es?color=blue)](./LICENSE) [![Build](https://github.com/connectrpc/connect-query-es/actions/workflows/ci.yaml/badge.svg?branch=main)](https://github.com/connectrpc/connect-query-es/actions/workflows/ci.yaml) [![NPM Version](https://img.shields.io/npm/v/@connectrpc/connect-query/latest?color=green&label=%40connectrpc%2Fconnect-query)](https://www.npmjs.com/package/@connectrpc/connect-query) [![NPM Version](https://img.shields.io/npm/v/@connectrpc/protoc-gen-connect-query/latest?color=green&label=%40connectrpc%2Fprotoc-gen-connect-query)](https://www.npmjs.com/package/@connectrpc/protoc-gen-connect-query)
 
-Connect-Query is an expansion pack for [TanStack Query](https://tanstack.com/query) (react-query), written in TypeScript and thoroughly tested.  It enables effortless communication with servers that speak the [Connect Protocol](https://connectrpc.com/docs/protocol).
+Connect-Query is an expansion pack for [TanStack Query](https://tanstack.com/query) (react-query), written in TypeScript and thoroughly tested. It enables effortless communication with servers that speak the [Connect Protocol](https://connectrpc.com/docs/protocol).
 
 - [Quickstart](#quickstart)
   - [Generated Code](#generated-code)
@@ -14,13 +15,15 @@ Connect-Query is an expansion pack for [TanStack Query](https://tanstack.com/que
   - [`createQueryService`](#createqueryservice)
   - [`TransportProvider`](#transportprovider)
   - [`useTransport`](#usetransport)
-  - [`UnaryHooks.createData`](#unaryhookscreatedata)
-  - [`UnaryHooks.createUseQueryOptions`](#unaryhookscreateusequeryoptions)
-  - [`UnaryHooks.getPartialQueryKey`](#unaryhooksgetpartialquerykey)
-  - [`UnaryHooks.getQueryKey`](#unaryhooksgetquerykey)
-  - [`UnaryHooks.methodInfo`](#unaryhooksmethodinfo)
-  - [`UnaryHooks.setQueryData`](#unaryhookssetquerydata)
-  - [`UnaryHooks.setQueriesData`](#unaryhookssetqueriesdata)
+  - [`UnaryFunctions.createData`](#unaryfunctionscreatedata)
+  - [`UnaryFunctions.createUseQueryOptions`](#unaryfunctionscreateusequeryoptions)
+  - [`UnaryFunctions.createUseInfiniteQueryOptions`](#unaryfunctionscreateuseinfinitequeryoptions)
+  - [`UnaryFunctions.createUseMutationOptions`](#unaryfunctionscreateusemutationoptions)
+  - [`UnaryFunctions.getPartialQueryKey`](#unaryfunctionsgetpartialquerykey)
+  - [`UnaryFunctions.getQueryKey`](#unaryfunctionsgetquerykey)
+  - [`UnaryFunctions.methodInfo`](#unaryfunctionsmethodinfo)
+  - [`UnaryFunctions.setQueryData`](#unaryfunctionssetquerydata)
+  - [`UnaryFunctions.setQueriesData`](#unaryfunctionssetqueriesdata)
   - [`UnaryHooks.useInfiniteQuery`](#unaryhooksuseinfinitequery)
   - [`UnaryHooks.useMutation`](#unaryhooksusemutation)
   - [`UnaryHooks.useQuery`](#unaryhooksusequery)
@@ -52,7 +55,7 @@ Note: If you are using something that doesn't automatically install peerDependen
 
 ### Usage
 
-Connect-Query will immediately feel familiar to you if you've used TanStack Query.  It provides a set of convenient helpers that you can pass to the same TanStack Query functions you're already using:
+Connect-Query will immediately feel familiar to you if you've used TanStack Query. It provides a set of convenient helpers that you can pass to the same TanStack Query functions you're already using:
 
 ```ts
 import { useQuery } from '@tanstack/react-query';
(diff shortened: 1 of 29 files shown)


README.md
```diff
@@ -495,60 +533,39 @@ Here is a high-level overview of how Connect-Query fits in with Connect-Web and
...
 
 ### Does this only work with React?
 
-You can use Connect-Query with any TanStack variant (React, Solid, Svelte, Vue).  However, since the hooks APIs like [`UnaryHooks.useQuery`](#unaryhooksusequery) and [`UnaryHooks.useMutation`](#unaryhooksusemutation) automatically infer `Transport` from React Context, these APIs will only work with React, as of now.  There is nothing else React specific in the Connect-Query codebase.  As we expand the scope of the project, we do hope to support all APIs on all TanStack Query variants.
-
-| Connect-Query API       | React              | Solid              | Svelte             | Vue             |
-| ----------------------- | ------------------ | ------------------ | ------------------ | ------------------ |
-| `createQueryService`    | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
```
timostamm, line 528: Is this - and the `UnaryHooks` -> `UnaryFunctions` rename - the only change to the documentation? (It's hard to tell because of the reformat.)

examples/react/basic/src/gen/eliza-BigIntService_connectquery.ts
```diff
@@ -41,9 +41,12 @@ export const BigIntService = {
...
 /**
  * @generated from rpc connectrpc.eliza.v1.BigIntService.Count
  */
-export const count = createQueryService({
-  service: BigIntService,
-}).count;
+export const count = {
+  ...queryService.count,
+  ...createHooks(queryService.count),
+};
```
timostamm, line 52: So these changes to protoc-gen-connect-query are not a breaking change, but we're adding new properties, if I understand correctly.

Do we explain anywhere what we add?

packages/connect-query/src/index.ts
```diff
@@ -12,23 +12,38 @@
...
 
+import type { Message } from "@bufbuild/protobuf";
+
+import type { UnaryHooks } from "./create-hooks";
+import type { UnaryFunctions } from "./create-unary-functions";
+
 export { createQueryService } from "./create-query-service";
 export type {
-  QueryHooks,
+  QueryFunctions,
```
timostamm, line 22: It looks like we're removing or changing several export, so we do make a breaking change here?

timostamm (commented):
I won't be much help with a review. There are a lot of changes in both plugins, the runtime, and the docs, but I don't understand what we do from the PR description. 

Maybe it would also be helpful for users to explain what we change and how it impacts them, for example with an added feature, or a breaking change that requires users to adapt how they use the API.

packages/connect-query/src/index.ts
paul-sachs: Yeah sorry, i forgot to put this PR into draft. I will add more details to the PR. There are some types that have breaking changes (though fairly light, with a replacement type that does the same). I'll update the description.

README.md
paul-sachs: Yeah going forward we'll be enforcing formatting on the markdown so that should stop. One of the changes is `UnaryHooks` -> `UnaryFunctions`. The added generated functions (`createUseInfiniteQueryOptions` and `createUseMutationOptions`) were also added to the docs.

I wanted to remove this section as I think it makes it a little more complicated looking than it actually is.

examples/react/basic/src/gen/eliza-ElizaService_connectquery.ts
```diff
@@ -93,20 +96,24 @@ export const ElizaService = {
   },
 } as const;
 
+const queryService = createQueryService({ service: ElizaService });
```
timostamm, line 99: I think this will generate uncompilable code with the following service:

```proto
service Locator {
  rpc QueryService(QueryServiceRequest) returns (QueryServiceResponse);
}
```

We have to avoid the collision. I think `$queryService` should be safe.

README.md
```diff
@@ -495,60 +533,39 @@ Here is a high-level overview of how Connect-Query fits in with Connect-Web and
...
+You can use Connect-Query with any TanStack variant (React, Solid, Svelte, Vue). All methods which are part of the `UnaryFunctions` type can be used by any framework. The only APIs which are React specific are the `TransportProvider`, and any APIs starting with `use`.
+
+> Tip: If you're a TanStack Query user that uses something other than React, we'd love to hear from you. Please reach out to us on the [Buf Slack](https://buf.build/links/slack).
 
 #### SolidJS Example
 
-Say, for example, you're using Solid together with TanStack Query's `useQuery` API.  In this case you can use `UnaryHooks.createUseQueryOptions` instead of `UnaryHooks.useQuery`.  The only difference is that `createUseQueryOptions` requires you to pass in a `Transport` because it is not a hook and hooks are where transport is automatically inferred.
+Say, for example, you're using Solid together with TanStack Query's `useQuery` API. In this case you can use `UnaryFunctions.createUseQueryOptions` instead of `UnaryHooks.useQuery`. The only difference is that `createUseQueryOptions` requires you to pass in a `Transport` because it is not a hook and hooks are where transport is automatically inferred.
```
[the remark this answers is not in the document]
paul-sachs: Yes, that is true. In general, this probably needs some tweaks.

examples/react/basic/src/gen/eliza-ElizaService_connectquery.ts
paul-sachs: Ooo nice catch. I'll make that modification.

packages/connect-query/src/create-unary-functions.ts
```diff
@@ -76,12 +75,12 @@ export interface UnaryHooks<I extends Message<I>, O extends Message<O>> {
    * The caveat being that if you go the route of using `createUseQueryOptions` you must provide transport.  You can get transport from the `useTransport` export.  If you cannot use hooks to retrieve transport, then look at the documentation for `TransportProvider` to learn more about how to use Connect-Web's createConnectTransport` or `createGrpcWebTransport`APIs.
    */
```
smaye81, line 76: Should this comment be updated based on the changes below? Seems like the transport is now optional but this comment says it must be provided.

packages/connect-query/src/create-unary-functions.ts
```diff
@@ -163,59 +162,43 @@ export interface UnaryHooks<I extends Message<I>, O extends Message<O>> {
...
-    { callOptions, getPlaceholderData, onError, transport },
+    // istanbul ignore next
+    { callOptions, getPlaceholderData, onError, transport } = {},
   ) => {
     const enabled = input !== disableQuery;
 
     assert(
-      transport !== undefined, // eslint-disable-line @typescript-eslint/no-unnecessary-condition -- yes, it's true that according to the types it should not be possible for a user to pass `undefined` for transport, but it's much nicer to catch them here if they do (as in, without TypeScript or in a insufficiently sound TypeScript configuration).
+      transport !== undefined,
       "createUseQueryOptions requires you to provide a Transport.  If you want automatic inference of Transport, try using the useQuery helper.",
```
smaye81, line 202: Is this condition still true also based on the above?

smaye81 (commented):
I'm not super familiar with the code, so I probably can't give an in-depth review either. But, I think I get the general approach here. Left a few small comments/questions.

packages/connect-query/src/create-unary-functions.ts
paul-sachs: Ah good catch. No, it's not technically required (if the service was created with a provided transport up front). But in most cases, it must be provided. I'll update the comment.

paul-sachs: I've updated this a little, though honestly I feel the fact this isn't just a mandatory field a little annoying. It's only this way since `createUnaryFunctions` itself takes an optional transport which this can fallback to.

smaye81 (approved):

timostamm (approved):
Thanks for the context in offline discussion, Paul. I think we still need to explain a bit better what we do:

- We make a breaking change here by changing the exports from @connectrpc/connect-query. Users might rely on them, so this needs to be clearly called out, and release must bump the minor version (a breaking change for 0.x versions).
- The code generated by protoc-gen-connect-query and protoc-gen-connect-query-react ("plugin API" is ambiguous) receives new methods. We should explain what they are for.
- If the refactoring regarding React is a user-facing change (giving users new features), we have to explain what the new feature is. If it's just future proofing, that is fine - usually it makes sense to do that separately from breaking changes and adding features, though. 

Can you update the PR description?

paul-sachs:
@timostamm thanks for the detailed response.

> * We make a breaking change here by changing the exports from @connectrpc/connect-query. Users might rely on them, so this needs to be clearly called out, and release must bump the minor version (a breaking change for 0.x versions).

Yeah, we'll definitely do a minor version bump after this is merged. I'll update the PR to draw more attention to it.

> * The code generated by protoc-gen-connect-query and protoc-gen-connect-query-react ("plugin API" is ambiguous) receives new methods. We should explain what they are for.

That makes sense, i'll add some explicit mention about what they're for.

> * If the refactoring regarding React is a user-facing change (giving users new features), we have to explain what the new feature is. If it's just future proofing, that is fine - usually it makes sense to do that separately from breaking changes and adding features, though.

The refactoring and added features are kind of a side effect of the future proofing rather than in addition to, we'd have to go out of our way in this case to prevent the features from leaking out. As mentioned, I'll add more details to the description and get this in.

PR #173, opened 2023-09-21
Generate proper service methods without hooks

paul-sachs:
The useXQuery methods cover these cases so we should only expose non-hooks through these exports.

The below API was never published so it's not a breaking change, just fixing generated `.ts` code to match what I had expected it to be.


timostamm (commented):
I don't understand what proper service methods are and what this changes. But I noticed that only .ts generated code is modified, not .js and .d.ts. Is this intentional?

paul-sachs:
@timostamm ah sorry, should have been more explicit. I also feel like I don't have good naming schemes for what each of these methods generate. Basically, there was a bug in my generated code where `const method = ...` was actually just the hooks (useQuery, useMutation, and useInfiniteQuery) which was accidental. It should have contained every function OTHER than those.

It doesn't contain generated js code because we actually never made the change to the react plugin to do the same thing the older plugin did. I will update that.

paul-sachs:
Actually, that is a much larger change. the `-react` plugin actually never generated proper js code, so I'm going to file a different PR to fix that.

timostamm:
Got it. Yeah, the terminology is a bit challenging here.

timostamm (approved):

PR #175, opened 2023-09-22
Fix js output

paul-sachs:
As part of prep for next release, fixing a few _react hook bugs and adding compatibility to generate js.

Also add missing transport from _react hook calls.


packages/protoc-gen-connect-query-react/src/generateTs.ts
```diff
@@ -41,6 +42,33 @@ const generateServiceFile =
     );
     f.preamble(protoFile);
 
+    /** @returns the printables associated with a param for a function. Handles the case where extension is js so types shouldn't be included.  */
+    function getParamWithTypes(
```
timostamm, line 46: My general recommendation would be to bite the bullet and use distinct functions for js, ts, and dts.

timostamm (approved):

packages/protoc-gen-connect-query-react/src/generateTs.ts
paul-sachs: yeah, it was convenient when they were 1:1 but now it's getting more and more clunky. Might have a followup. I just realized now that the output .ts is also not running through a type check process which is weird and i'd like the fix that.

PR #176, opened 2023-09-22
Split js/ts generator

paul-sachs:
- Added checks to run tsc on generated output
- Split js/ts generator

This should be the final PR before a PR to move to recommending this plugin over the old one


packages/protoc-gen-connect-query-react/package.json
```diff
@@ -21,13 +21,16 @@
   "scripts": {
     "clean": "rm -rf dist",
     "build": "pnpm clean && tsc --project tsconfig.build.json",
-    "format": "prettier . --write --ignore-path ./.eslintignore && eslint . --fix && license-header",
+    "format": "prettier . --write --ignore-path ./.eslintignore && eslint . --fix && license-header && tsc --noEmit --project tsconfig.json",
```
paul-sachs, line 24: This change, plus the addition of the `ts_nocheck` flag sent to the buf plugin make it so we can run tsc to validate the resulting content.

timostamm (commented):
LGTM. Just needs a prettier run, I think.

timostamm (approved):

PR #177, opened 2023-09-25
Release 0.5.1

paul-sachs:
## What's Changed
* Extract React specific dependencies into single file by @paul-sachs in https://github.com/connectrpc/connect-query-es/pull/169
* Split js/ts generator by @paul-sachs in https://github.com/connectrpc/connect-query-es/pull/176

**Full Changelog**: https://github.com/connectrpc/connect-query-es/compare/v0.4.4...v0.5.1

#### Breaking changes
- Removed `UnaryHooks` type and replaced with `UnaryFunctionsWithHooks`
- Updated `createUnaryHooks` to only return the hooks and not other functions
 
##### createUnaryFunctions

Introduced `createUnaryFunctions` which can be used with the new `createUnaryHooks` to do the same thing as the old `createUnaryHooks`. The reasoning behind this change is to prep the core library, `connect-query`, to be able to used separately from React itself and this is just the first step in that direction.

#### New functions

Each service method now generates two new functions associated to the method:

##### method.createUseInfiniteQueryOptions

This method creates options that will pass to `useInfiniteQuery`. It is identical to `method.useInfiniteQuery` except it needs to be passed a transport directly.

##### createUseMutationOptions

This method creates options that will pass to `useMutation`. It is identical to `method.useMutation` except it needs to be passed a transport directly.

We added these methods to better future-proof the code for incoming changes to React (React Server Components, etc).  

## Migration

- Replace any usages of the type `UnaryHooks` with `UnaryFunctionsWithHooks`
- Find any usages of `createUnaryHooks` and use a combination of `createUnaryHooks` and `createUnaryFunctions`

**Before**
```tsx
const say = createUnaryHooks(ExampleService);
```

**After**
```tsx
const say = {
  ...createUnaryHooks(ExampleService),
  ...createUnaryFunctions(ExampleService),
}
```


timostamm (approved):

paul-sachs:
Messed up publishing of 0.5.0 (changed branch and forgot to rerun build). Bumping to 0.5.1 and publishing again.

PR #179, opened 2023-09-27
Update to react-query@5

paul-sachs:
Though most of the types have stayed the same, there are a few breaking changes to be aware about:

- Removed `onError` passed to `useQuery`, this aligns with the removed event handlers in [@tanstack/query](https://tanstack.com/query/v5/docs/react/guides/migrating-to-v5#callbacks-on-usequery-and-queryobserver-have-been-removed)
- `useInfiniteQuery` now requires the initial page param to be defined in the input because `initialPageParam` is [required](https://tanstack.com/query/v5/docs/react/guides/migrating-to-v5#infinite-queries-now-need-a-initialpageparam) 

This will require a bump to minor version.


smaye81 (approved):

packages/connect-query/package.json
```diff
@@ -52,7 +52,7 @@
   "peerDependencies": {
     "@bufbuild/protobuf": "^1.3.3",
     "@connectrpc/connect": "^1.1.2",
-    "@tanstack/react-query": "4.x",
+    "@tanstack/react-query": "5.0.0 < 6.0.0",
```
timostamm, line 55: This constraint won't match version `5.0.5`, the current latest.

paul-sachs: Oh man, yeah I totally missed that. I'm actually weirded out that it only matches 5.0.0 🤔 , oh well, I'll update it to `5 < 6` which seems to match fine (according to https://semver.npmjs.com/)

paul-sachs: Strike that, 5 < 6 doesn't work the way I thought either... Maybe just `5.x` is just fine.

timostamm: We can simply use `^5.0.0` here. I recommend this tool: https://github.com/jubianchi/semver-check

PR #184, opened 2023-09-29
Integrate `arethetypeswrong`

### LINKED ISSUE #182: Verify our exports with arethetypeswrong
https://arethetypeswrong.github.io/ verifies type exports of NPM packages with the version resolution algorithms implemented in TypeScript. 

It's available as a CLI, and we should use it here to guard against regressions. For reference, see https://github.com/connectrpc/connect-es/pull/838.

smaye81:
Fixes #182 

This integrates the `arethetypeswrong` CLI to the codebase and runs it in CI, fixing any type issues that existed.

### DIFF
--- a/packages/connect-query/package.json
+++ b/packages/connect-query/package.json
@@ -11,14 +11,12 @@
   "type": "module",
   "exports": {
     ".": {
-      "types": "./dist/index.d.ts",
-      "import": "./dist/index.js",
-      "require": "./dist/index.cjs"
+      "module": "./dist/index.js",
+      "require": "./dist/index.cjs",
+      "import": "./dist/index.js"
     }
   },
   "main": "./dist/index.cjs",
-  "module": "./dist/index.js",
-  "types": "./dist/index.d.ts",
   "files": [
     "dist/**"
   ],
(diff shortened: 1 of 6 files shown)


paul-sachs (approved):

packages/connect-query/package.json
```diff
@@ -11,14 +11,12 @@
   "type": "module",
   "exports": {
     ".": {
-      "types": "./dist/index.d.ts",
-      "import": "./dist/index.js",
-      "require": "./dist/index.cjs"
+      "module": "./dist/index.js",
```
timostamm, line 14: I suggest we merge this first, then merge it into https://github.com/connectrpc/connect-query-es/pull/180.
Feel free to commandeer that PR.

turbo.json
```diff
@@ -13,6 +13,9 @@
       "dependsOn": ["generate"],
       "cache": false
     },
+    "attw": {
+      "dependsOn": ["format"]
```
timostamm, line 17: `attw` runs `npm pack`, so it needs the build artifacts. It seems very odd to rely on the build being run through two levels of dependencies. Can't we simply do

```suggestion
      "dependsOn": ["build"]
```

here?

timostamm (approved):

PR #185, opened 2023-09-29
Added FAQ question about prefetching

### LINKED ISSUE #117: Simplify usage with prefetchQuery
The generated hooks doesn't provide a very ergonomic way to interact with `queryClient.prefetchQuery`. See slack thread [here](https://bufbuild.slack.com/archives/CRZ680FUH/p1692220929043539).

If you want to build a function to prefetch the query, you need to use a hook to initialize the options and this limits the places we can do that. This problem is somewhat related to #55 , wherein we want methods that don't depend on context to be usable.

paul-sachs:
Add docs about prefetching. 

Also adds tschecks for generated files in example.

Fixes #117


timostamm (approved):
LGTM. 

Will need to update the mirror here as well: https://github.com/connectrpc/connectrpc.com

PR #186, opened 2023-09-29
Migrate to Node16 resolution

paul-sachs:


packages/connect-query/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "./tsconfig.build.json",
   "compilerOptions": {
     "rootDir": ".",
+    "noEmit": true,
```
timostamm, line 5: Don't we want to emit here? 🤔

packages/protoc-gen-connect-query/tsconfig.build.json
```diff
@@ -2,8 +2,10 @@
   "files": ["src/protoc-gen-connect-query-plugin.ts"],
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "outDir": "dist/cjs",
+    "moduleResolution": "node",
     "module": "CommonJS",
-    "resolveJsonModule": true
+    "outDir": "dist/cjs",
+    "resolveJsonModule": true,
+    "skipLibCheck": true
```
timostamm, line 9: Do we really need to skip lib check with "node" version resolution?

packages/connect-query/src/stable-hash.js
```diff
@@ -0,0 +1,26 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import stableHash from "stable-hash";
+
+/**
+ * Convert stable-hash to a typed API. This is a necessary workaround because
+ * stable-hash does not export a type for its default export that is compatible
+ * with Node16 resolution.
```
timostamm, line 20: attw reports "no types" for Node16 from ESM. https://arethetypeswrong.github.io/?p=stable-hash%400.0.3

That's a nice workaround, but is the JS file really part of `dist` after a build?

packages/connect-query/package.json
```diff
@@ -11,8 +11,8 @@
   "scripts": {
     "clean": "rm -rf ./dist/*",
     "build": "npm run build:cjs && npm run build:esm && npm run build:proxy",
-    "build:cjs": "tsc --project tsconfig.build.json --module commonjs --outDir ./dist/cjs --declaration --declarationDir ./dist/cjs && echo >./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
-    "build:esm": "tsc --project tsconfig.build.json --module ES2015 --verbatimModuleSyntax --outDir ./dist/esm --declaration --declarationDir ./dist/esm",
+    "build:cjs": "tsc --project tsconfig.build.json --module commonjs --moduleResolution node --outDir ./dist/cjs --declaration --declarationDir ./dist/cjs && echo >./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
+    "build:esm": "tsc --project tsconfig.build.json --module Node16 --verbatimModuleSyntax --outDir ./dist/esm --declaration --declarationDir ./dist/esm",
```
timostamm, line 15: Can we remove `--module Node16` here, since it's already set in tsconfig? I think we can also add verbatimModuleSyntax to tsconfig, and use `--verbatimModuleSyntax false` in `build:cjs`. The idea is that we do a standard compile for ESM, and make adjustments only for CJS.

packages/protoc-gen-connect-query/tsconfig.build.json
timostamm: We're basically not making any change for the plugin because of https://github.com/bufbuild/protobuf-es/issues/586, which should not be an issue for users at all because all Node.js versions can run the CJS binary just fine, right?

paul-sachs: Yeah, my intention was to keep this one the same, probably just got lost in the shuffle back and forth. I'll keep this CJS entirely.

packages/connect-query/tsconfig.json
paul-sachs: Hmm, I think this might be an unnecessary change. Might have been switching contexts too much.

paul-sachs: Ah i see why i added this. It was a little weird, but there's an error in the this file (when looking at it in vscode): 

```
Cannot write file '/Users/paulsachs/code/connect-query-es/packages/connect-query/src/stable-hash.js' because it would overwrite input file.
```

Since this file is pretty much purely for editors to be able to default to something, I don't think adding `noEmit` harms anything but I agree, it is a little confusing.

packages/connect-query/src/stable-hash.js
paul-sachs: Yep, tsc includes that file during a compile.

packages/protoc-gen-connect-query/tsconfig.build.json
paul-sachs: And it turns out, `skipLibCheck` is not required. It probably got added when I was messing with other settings. Removed.

packages/connect-query/tsconfig.json
paul-sachs: Oh, i guess tests also use this tsconfig.

paul-sachs:
@timostamm @smaye81 resurrecting this PR to migrate to node16.

packages/connect-query/tsconfig.build.json
```diff
@@ -1,7 +1,11 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "verbatimModuleSyntax": true,
```
timostamm, line 5: I think it would be best to enable verbatimModuleSyntax in tsconfig.base.json. This show errors in the IDE instead of later in the build.

tsconfig.base.json
```diff
@@ -14,7 +16,7 @@
     "noImplicitOverride": true,
 
     // We need node's module resolution, so we do not have to skip lib checks
-    "moduleResolution": "Node",
+    "moduleResolution": "Node16",
```
timostamm, line 19: Can we also add `"module": "Node16"` here? I don't even know from the top of my head whether it's automatically set. Setting it explicitly seems a bit more clear.

packages/connect-query/package.json
```diff
@@ -11,8 +11,8 @@
   "scripts": {
     "clean": "rm -rf ./dist/*",
     "build": "npm run build:cjs && npm run build:esm && npm run build:proxy",
-    "build:cjs": "tsc --project tsconfig.build.json --module commonjs --outDir ./dist/cjs --declaration --declarationDir ./dist/cjs && echo >./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
-    "build:esm": "tsc --project tsconfig.build.json --module ES2015 --verbatimModuleSyntax --outDir ./dist/esm --declaration --declarationDir ./dist/esm",
+    "build:cjs": "tsc --project tsconfig.build.json --module commonjs --moduleResolution node --verbatimModuleSyntax false --outDir ./dist/cjs --declarationDir ./dist/cjs && echo >./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
```
timostamm, line 14:
```suggestion
    "build:cjs": "tsc --project tsconfig.build.json --module commonjs --moduleResolution node10 --verbatimModuleSyntax false --outDir ./dist/cjs --declarationDir ./dist/cjs && echo >./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
```

From the TS docs:

> 'node10' (previously called 'node') for Node.js versions older than v10, which only support CommonJS require. You probably won’t need to use node10 in modern code.

timostamm (commented):
@paul-sachs, we just switched protobuf-es to build with Node16 module resolution, see https://github.com/bufbuild/protobuf-es/pull/754. It's not released yet, and connect-es has not been updated yet (see https://github.com/connectrpc/connect-es/issues/887).

The changes here look good to me. Just left a couple of comments.

tsconfig.base.json
paul-sachs: Actually it was already there, just above a bit. I'll move it down to be with this though.

paul-sachs: Ah i see, no i got confused.

paul-sachs: Anyways, it's here now.

PR #187, opened 2023-09-30
Update docs with accurate createQueryService docs

paul-sachs:
I had missed these docs in the during the update to 5.0.


timostamm (approved):
Does https://github.com/connectrpc/connectrpc.com need an update as well?

PR #211, opened 2023-10-05
Fix missing implicit types

### LINKED ISSUE #209: disableQuery unique symbol stopping typescript declaration
We use this lib to publish a client to npm that our different code bases can share. However, there seems to be issues I can't figure out surrounding `unique symbol`'s and typescript declaration. 

I noticed you have an exception for this case in your eslint, and it is for a test file to consume the generated lib. 
Do you know of a work around beyond having an unused `disableQuery` import everywhere I'd want to use the generated code?

mckelveygreg:
We have really appreciated this lib, and have been using it since it came out! I think I found a typing fix to make it even more usable.

When trying to bundle up the generated code with `tsc --declaration=true`, ts couldn't figure out the `disableQuery` type and the `BaseInfiniteQueryOptions` type. This is able to be resolved if the exported queries are explicitly typed.

This also will resolve #209 🎉 

Let me know if there is anything else I can do to this PR!


timostamm (commented):
Hey Greg, I believe this is the correct fix, thanks for the contribution!

We did the same in protobuf-es in https://github.com/bufbuild/protobuf-es/pull/398. 

I'll defer to @paul-sachs regarding any implementation details.

packages/protoc-gen-connect-query/src/generateTs.ts
```diff
@@ -74,13 +75,22 @@ const generateServiceFile =
     );
     f.print();
 
-    
+    const unaryFunctionType = (method: DescMethod) => [f.import('UnaryFunctions', '@connectrpc/connect-query'), `<${method.input.name}, ${method.output.name}>`]
```
paul-sachs, line 78: @mckelveygreg This looks good, though I think we can simplify by using the helper type `UnaryFunctionsWithHooks` instead of `UnaryFunctions<...> & UnaryHooks<...>`.

mckelveygreg: 😅 Ha, yeah that is much more straight forward!
Thanks!

paul-sachs (approved):

PR #214, opened 2023-10-06
Run tsc against generated code

paul-sachs:
This makes sure that the generated code doesn't have any finicky TS errors. This change would have caught #209 

Also remove workspace designation on example so it can be independently cloned.


packages/protoc-gen-connect-query/package.json
```diff
@@ -22,13 +22,17 @@
     "clean": "rm -rf dist",
     "build": "pnpm clean && tsc --project tsconfig.build.json",
     "format": "prettier . --write --ignore-path ./.eslintignore && eslint . --fix && license-header",
-    "generate": "rm -rf snapshots/gen && cd snapshots && buf generate buf.build/connectrpc/eliza:8b8b971d6fde4dc8ba5d96f9fda7d53c"
+    "generate": "rm -rf snapshots/gen && cd snapshots && buf generate buf.build/connectrpc/eliza:8b8b971d6fde4dc8ba5d96f9fda7d53c",
+    "test": "tsc --noEmit"
```
timostamm, line 26: Are you sure that `tsc --noEmit` would have caught the issue?

In https://github.com/bufbuild/protobuf-es/pull/398, we actually had to emit declaration files to fail compilation.

paul-sachs: It caught it during my tests anyways. As long as I removed the no-check flag. I'll double check to make sure I'm not imagining it

timostamm (approved):

PR #219, opened 2023-10-16
Rename packages

paul-sachs:
Renaming `connect-query` to `connect-react-query` and `protoc-gen-connect-query` to `protoc-gen-connect-react-query`. This makes it clear that these packages still very much depend on running in a react environment.

These changes will be published as v1


examples/react/basic/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@connectrpc/connect-query-example-react-basic",
+  "name": "@connectrpc/connect-react-query-example-react-basic",
```
timostamm, line 2:
```suggestion
  "name": "@connectrpc/connect-react-query-example-basic",
```

packages/protoc-gen-connect-react-query/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@connectrpc/protoc-gen-connect-query",
+  "name": "@connectrpc/protoc-gen-connect-react-query",
```
timostamm, line 2: Shouldn't we rename the npm binary as well?

timostamm (commented):
I think we're missing a couple of bits, see comments below. Should the extension for generated files remain `_connectquery.ts`?

packages/protoc-gen-connect-react-query/package.json
paul-sachs: Ah right, yeah, good catch.

paul-sachs:
> Should the extension for generated files remain `_connectquery.ts`?

That's a good question. I think it stands to reason that it should be `_connectreactquery.ts`, even it makes me hurt a little inside.

paul-sachs:
@timostamm how do you feel about using `_cq_react.ts` as a file ending instead of `_connectreactquery.ts` postfixes? I kinda like keeping the postfix short since we're already prefixed with the service and the method.

paul-sachs:
Closing this until we can discuss what renaming would cost us.

PR #223, opened 2023-10-23
Release 0.6.0

paul-sachs:
## What's Changed
* Update to react-query@5 by @paul-sachs in https://github.com/connectrpc/connect-query-es/pull/179

### Update to react-query@5

This update contains some breaking changes including, but not limited to:
- Removed onError passed to useQuery, this aligns with the removed event handlers in [@tanstack/query](https://tanstack.com/query/v5/docs/react/guides/migrating-to-v5#callbacks-on-usequery-and-queryobserver-have-been-removed)
- useInfiniteQuery now requires the initial page param to be defined in the input because initialPageParam is [required](https://tanstack.com/query/v5/docs/react/guides/migrating-to-v5#infinite-queries-now-need-a-initialpageparam)

**Full Changelog**: https://github.com/connectrpc/connect-query-es/compare/v0.5.3...v0.6.0


examples/react/basic/package.json
```diff
@@ -16,7 +16,7 @@
     "@bufbuild/protobuf": "^1.3.3",
     "@bufbuild/protoc-gen-es": "^1.3.3",
     "@connectrpc/connect": "^1.1.2",
-    "@connectrpc/connect-query": "workspace:*",
+    "@connectrpc/connect-query": "0.5.1",
```
timostamm, line 19: This should be 0.6.0, I think. 

Can we simply use `*` as the version constraint in the example? Definitely not ideal, but perhaps better than having to manually update the field for a a version bump.

examples/react/basic/src/gen/eliza-BigIntService_connectquery.ts
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// @generated by protoc-gen-connect-query v0.5.3 with parameter "target=ts,import_extension=none,ts_nocheck=false"
+// @generated by protoc-gen-connect-query v0.5.1 with parameter "target=ts,import_extension=none,ts_nocheck=false"
```
timostamm, line 15: I'd expect `v0.6.0` here.

examples/react/basic/package.json
paul-sachs: Yeah that's kind the sideeffect of having an example that points to a real release. Without the actual release being done, we can't point to most recent here.

examples/react/basic/src/gen/eliza-BigIntService_connectquery.ts
paul-sachs: Yeah without the release, we can't reference 0.6.0. I'll switch back to workspace to bring this back up to 0.6.0.

examples/react/basic/package.json
paul-sachs: Returning to workspace for now. Will need to think on a better way to keep these in sync without using a workspace reference.

timostamm: Ah, so pnpm only links the `workspace:*` reference? That's a bummer.

timostamm (approved):

examples/react/basic/package.json
paul-sachs: Sorta bummer but also nice and predictable, it prevents accidentally using one over the other.

PR #224, opened 2023-10-23
V1 API: Wrapping react-query

### LINKED ISSUE #220: Proposal: API improvements
The user for facing API today looks like:
```ts
import { useQuery } from '@tanstack/react-query';
import { example } from 'your-generated-code/example-ExampleService_connectquery';

export const Example: FC = () => {
  const { data } = useQuery(example.useQuery({}));
  return <div>{data}</div>;
};
```
The generated code is as follows:
```ts
import {
  createQueryService,
  createUnaryHooks,
} from "@connectrpc/connect-query";
import { MethodKind } from "@bufbuild/protobuf";
import { ExampleRequest, ExampleResponse } from "./example_pb.js";

export const typeName = "your.company.com.example.v1.ExampleService";

export const ExampleService = {
  methods: {
    example: {
      name: "Example",
      kind: MethodKind.Unary,
      I: ExampleRequest,
      O: ExampleResponse,
    },
  },
  typeName,
} as const;

const $queryService = createQueryService({ service: ExampleService });

export const say = {
  ...$queryService.example,
  ...createUnaryHooks($queryService.example),
};
```

### Issues with the above API and generated code:
- The use of `useQuery` twice is verbose and confusing, if one has to set other fields that the query APIs accept it gets even more verbose with a slightly different looking call: `useQuery({example.useQuery({})..., extra: value})`
- It is not very apparent what values are set by the generated hook and which of them are safe to override.
- Query APIs are endpoint oriented and as such the generated code correctly exports a unique symbol for each rpc. However, the way the code is generated today is not ideal for tree-shaking as it will include the entire service object (including all other methods). Not only that but the entire runtime is included even if only one of the hooks is used.
- For contributors to add support for other frameworks the example set by the react approach is to write a framework specific plugin and accompanying runtime package. This is a barrier for new contributors as they have to know about protobuf plugins and how to write one.
- Probably a consequence of how the code is generated, the API surface of the runtime package is more than ideal. 
- There are multiple ways to create the query hooks. One can use the generated code or they can simple use the generated code by the connect plugin and use the runtime to create their own hooks.
- Mocking the API requires them to use the service definition, this is different from how users use the hooks.

### Possible solution
I propose a framework agnostic plugin accompanied by a new react specific API that wraps the original hooks. The generated code will look like:

```ts
import { MethodKind } from "@bufbuild/protobuf";
import { ExampleRequest, ExampleResponse } from "./example_pb.js";

export const typeName = "your.company.com.example.v1.ExampleService";

export const example = {
  localName: "example",
  name: "Example",
  kind: MethodKind.Unary,
  I: ExampleRequest,
  O: ExampleResponse,
  service: { typeName },
} as const;
```
This solves the tree-shaking problem by just generating method specific definitions of the rpcs. This is also framework agnostic as it only generates the rpc definitions without any imports of query APIs.

The runtime API will take the form:
```ts
import { useQuery } from '@connectrpc/connect-react-query';
import { example } from 'your-generated-code/example-ExampleService_connectquery';

export const Example: FC = () => {
  const { data } = useQuery(example, {});
  return <div>{data}</div>;
};
```
In this API, the second parameter is for the request, the optional third is to provide additional parameters to the underlying Hooks API. This way we can only expose safe to override properties. This will also make wrapping the hooks to modify the types easier (https://github.com/connectrpc/connect-es/issues/694#issuecomment-1677929961). We should also expose supplementary API for things like calculating query keys and other bits that help us construct the hooks. 

Along with this we should expose a mocking wrapper similar to `createRouterTransport` that is rpc oriented where users can import the connect-query generated constants to mock rpcs. 

With this structure contributors only need to create the runtime package and not bother about the plugin. 

### Implementation

This is a breaking change and a significant shift from our current API. I am not entirely sure if we can write a migrator but even if we did it won't be trivial. 

One option is to hold off on #219 and #179 and release a parallel package for Tanstack v5. We continue to support the current version for Tanstack v4 and users wanting to upgrade to Tanstack v5 can update to the new react specific package. This gives us some time to work on the migrator and at the same time new users will start using the new APIs.

@paul-sachs @timostamm thoughts?

paul-sachs:
Introducing a whole new API for connect-query. This API ties itself more tightly with the fantastic `@tanstack/react-query` in order to improve developer experience and reduce bundle size.

At the core of this change is the new `MethodUnaryDescriptor` type, which is essentially just a self contained description of a service method. The new code generator just outputs one of these per method and those are passed to the hooks/methods. This new additional type brings a huge level of flexibility about how we write additional methods, and it's trivial to write your own hooks/methods around these simple types.

## Example usage

```tsx
import { useQuery } from "@connectrpc/connect-react-query";
import { say } from "./gen/eliza-ElizaService_connectquery";

...

const { data } = useQuery(say);
```

## Reasoning

There are a number of reasons we've decided to make this major breaking change.

1. The API surface is much smaller, leading to less confusion about how to use each method.
2. Smaller core code size and more modular organization leads to better tree shaking (since each generated method doesn't come along with a generated hook).
3. Package names are now explicit about their dependencies, making it easier to develop other packages based on these packages/generated code.
4. More tightly integrating with `@tanstack/react-query@5` provides better usage of Suspense versions of the API.
5. New generated code is easier to expand and build your own custom hooks for.

## Migration

The migration process is easy:

### Before

```tsx
import { getUserOrganization } from "@apigen/org/alpha/registry/v1alpha1/organization-OrganizationService_connectquery";
import { useQuery } from "@tanstack/react-query";

...

const getUserOrganizationQuery = useQuery({
    ...getUserOrganization.useQuery({
      userId: currentUser?.id,
      organizationId,
    }),
    useErrorBoundary: false,
    enabled: currentUser !== null,
});
```

### After

```tsx
import { getUserOrganization } from "@apigen/org/alpha/registry/v1alpha1/organization-OrganizationService_connectquery";
import { useQuery } from "@connectrpc/connect-react-query";

...

const getUserOrganizationQuery = useQuery(getUserOrganization, {
      userId: currentUser?.id,
      organizationId,
    },
    {
      useErrorBoundary: false,
      enabled: currentUser !== null
    }
});
```

Fixes #220


packages/connect-react-query/src/use-query.ts
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Query the method provided. Maps to useQuery on tanstack/react-query
+ *
+ * @param methodSig
+ * @returns
+ */
+export function useQuery<I extends Message<I>, O extends Message<O>>(
+  methodSig: MethodUnaryDescriptor<I, O>,
+  input?: DisableQuery | PartialMessage<I> | undefined,
```
srikrsna-buf, line 46: Should this be optional? If so why is this also `undefined` and why is the input for `useSuspenseQuery` not optional?

packages/connect-react-query/src/method-unary-descriptor.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/** Defines a standalone method and associated service  */
+export type MethodUnaryDescriptor<
+  I extends Message<I>,
+  O extends Message<O>,
+> = MethodInfoUnary<I, O> & {
+  localName: string;
+  service: {
+    typeName: string;
+  };
```
srikrsna-buf, line 25:
```suggestion
  readonly service: Omit<ServiceType, "methods">;
```
Because we may want other features like options in the future.

srikrsna-buf (commented):
API looks great! Left some questions

examples/react/basic/src/gen/eliza-ElizaService_connectquery.ts
```diff
@@ -96,24 +91,34 @@ export const ElizaService = {
...
+export const sayAgain = {
+  localName: "sayAgain",
+  name: "SayAgain",
+  kind: MethodKind.Unary,
+  I: SayRequest,
+  O: SayResponse,
+  service: {
+    typeName: "connectrpc.eliza.v1.ElizaService",
+  },
+} as const;
```
timostamm, line 124: We generate this line further up:

```ts
export const typeName = "connectrpc.eliza.v1.ElizaService";
```

We should either use this const, or remove it. 

If we want to use the const, we have to avoid name conflicts. Currently, an `rpc TypeName(TypeNameRequest) returns (TypeNameResponse)` will generate invalid code.

On a separate note: We generate the full `ElizaService` above. Ultimately, it would be great to have a self-describing type for methods as a first-class primitive, and have ConnectRouter.rpc() accept it, for example:

```diff
const t = createRouterTransport(router => {
-  router.rpc(say.service, say, (req) => {
+  router.rpc(say, (req) => {
    return new SayResponse({sentence: `you said: ${req.sentence}`});
  })
});
```

If we can settle on the shape, I'm in favor of modifying the router transport to accept this type as well.

packages/protoc-gen-connect-query/src/generateTs.ts
```diff
@@ -67,30 +67,23 @@ const generateServiceFile =
     f.print("}", isTs ? " as const" : "", ";");
```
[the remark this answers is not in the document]
paul-sachs: I left this for ease of writing tests. Ideally, we'd remove this before we actually merge it, assuming we're happy with the method descriptor.

packages/connect-react-query/src/index.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export type {
+  ConnectQueryKey,
+  ConnectPartialQueryKey,
+} from "./connect-query-key.js";
+export { disableQuery } from "./utils.js";
+export { useTransport, TransportProvider } from "./use-transport.js";
+export type {
+  UseInfiniteQueryOptions,
+  createUseInfiniteQueryOptions,
```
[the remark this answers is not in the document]
paul-sachs: Yes, these are needed when used with `useQueries({ queries: [ createUseQueryOptions(...) ] })` in order to define parallel queries. I thought perhaps we could define our own `useQueries()` but that would mean it would be impossible to mix connect queries with other external queries.

packages/connect-react-query/src/use-query.ts
paul-sachs: I kept this optional so you can write:

```tsx
const { data } = useQuery(say);
```

instead of:

```tsx
const { data } = useQuery(say, {});
```

It's only inconsistent because I messed up. I will make it consistent. Actually not sure why it's a combination of `?` and `undefined`. That was a little bit of a hold over from the original library and I can't say I ever understood why.

packages/connect-react-query/src/create-use-query-options.ts
```diff
@@ -0,0 +1,142 @@
...
+> = ConnectQueryOptions &
+  Omit<
+    TSUseQueryOptions<O, ConnectError, O, ConnectQueryKey<I>>,
+    "queryFn" | "queryKey"
+  >;
+
+/**
+ * Options for useQuery
+ */
+export type UseSuspenseQueryOptions<
```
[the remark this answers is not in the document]
paul-sachs: Good idea, this naming was getting out of hand.

examples/react/basic/src/gen/eliza-ElizaService_connectquery.ts
paul-sachs: Yes, absolutely. If we're happy with the method declaration of `MethodUnaryDescriptor`, then we can propagate that back into `ConnectRouter` and I can delete this extra stuff, which would be great.

packages/connect-react-query/src/connect-query-key.ts
```diff
@@ -57,12 +53,16 @@ export type ConnectPartialQueryKey = [
...
  * @see ConnectQueryKey for information on the components of Connect-Query's keys.
  */
-export const makeConnectQueryKeyGetter =
-  (typeName: ServiceType["typeName"], methodInfoName: MethodInfo["name"]) =>
-  <I extends Message<I>>(
-    input?: DisableQuery | PartialMessage<I> | undefined,
-  ): ConnectQueryKey<I> => [
-    typeName,
-    methodInfoName,
+export function createConnectQueryKey<
```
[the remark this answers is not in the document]
paul-sachs: Ah weird, I thought this was exported. Absolutely, I'll add it.

examples/react/basic/src/gen/eliza-ElizaService_connectquery.ts
paul-sachs: I can submit a PR to `@bufbuild/protobuf` and `@connectrpc/connect` to contain whatever those represent. The only question I have is the use of `localName`, I'm not sure it needs to exist in the definition, since it really can be generated if anyone needs it. Thoughts @srikrsna-buf ?

paul-sachs: Went ahead and implemented it here: https://github.com/bufbuild/protobuf-es/pull/594. I'm all ears for what we'd want to name it though, `MethodType` isn't my favourite but it felt consistent with `ServiceType`.

timostamm: 👍  let's discuss it over there.

timostamm: This was discussed in https://github.com/bufbuild/protobuf-es/pull/594, and implemented as an overload for `ConnectRouter.rpc` in https://github.com/connectrpc/connect-es/pull/884. Not released yet, so the code cannot be updated yet.

srikrsna-buf:
@tannerlinsley It'd be great to get your inputs on the new API. I think you also had feedback regarding what the hooks accept as input for the rpc requests.

paul-sachs:
@srikrsna-buf @timostamm Alright, i've updated the readme for a pass at documentation (update to doc page will obviously need to wait). I wanted to gauge a few ideas about release.

- `protoc-gen-connect-query`: This plugin is now not even really connect-query related and is really just spitting out generic method descriptors. We can keep it like that for now, but if we create any other tools based on those descriptors, it probably makes sense to move this plugin elsewhere and rename it. We can obviously punt that decision till later but I wanted to raise it now.
- `connect-react-query`: really just a rename and a v1 all at the same time. I thought it made sense to do the rename here as well as the v1 since both can be considered breaking so might as well make it a single breaking change.

srikrsna-buf (commented):
Looks great! Just had one suggestion:

I think we should expose functions to call an rpc (unary to begin with) that accept a transport and the request in a type safe way. Then users will never need to interact with `connect-web` and more importantly `protoc-gen-connect-es` and can directly call rpcs. This is will more or less look like the unary query function.

If we expose this function then we can also un export the `create*` functions reducing the API surface significantly.

timostamm (commented):
Thanks for the updates to the docs!

Krishna's suggestion seems to be worthwhile to me.

paul-sachs:
@srikrsna-buf hmm, I see what you mean. So instead of having something link:

```ts
const [result1, result2] = useQueries({
  queries: [
    createUseQueryOptions(serviceMethod1, input1, options1),
    createUseQueryOptions(serviceMethod2, input2, options2)
  ]
});
```

We'd just document/encourage the following:

```ts
const [result1, result2] = useQueries({
  queries: [
    { queryKey: createConnectQueryKey(serviceMethod1, input1), queryFn: createConnectQueryFn(serviceMethod1, input1, options1) },
    { queryKey: createConnectQueryKey(serviceMethod2, input2), queryFn: createConnectQueryFn(serviceMethod2, input2, options2) },
  ]
});
```

I find that a little more error prone myself (due to potential mismatches between queryKey and queryFn) but it does admittedly reduce the API surface layer. We'd still need another function for infiniteQueries since the key generated for that is actually slightly different than our default key (since pageParam impacts the key).

Perhaps we can simplify the exported options a little bit though.The difference between suspense/non-suspense APIs are entirely within the options they take, to make it consistent with other APIs (like useQuery/useSuspenseQuery). Perhaps a function similar to our old API that just returns `queryKey` and `queryFn` would suffice, allowing the user to add other options as they desire (onError, onSuccess, etc) by just spreading the result. That would at least remove two APIs , `createUseSuspenseInfiniteQueryOptions` and `createUseSuspenseQueryOptions`.

paul-sachs:
@srikrsna-buf @timostamm updated docs and API to remove the dedicated `createUseSuspense*Options` APIs. I think this is as tiny an API I can get while still allowing for full access to typesafe react-query.

timostamm (commented):
Really nice, let's get this merged! 

But the PR title "Experimental API suggestion" is not descriptive. The title needs to summarize what this does. 

Also:

1. When this is released, we need to explain crystal clear what is changing, why, and what users need to do. IMO it's best to do that right in the PR description, but I'm fine doing it just in the release notes if you prefer.

2. Does it make sense to rename the package `@connectrpc/connect-query` to `@connectrpc/connect-react-query` in this PR? Splitting it out into a separate change would make it quite a bite easier to follow what's going on.

paul-sachs:
@timostamm i will definitely update the pr description to layout all the changes and why. Plus some migration examples. 

As for #2, I was originally thinking that this plug-in was no longer react specific so it could be used for any version of connect-query but since that doesn't exist yet, perhaps that's optimizing for a later case that's not important. I would be down with renaming the plug-in.

timostamm:
I can't follow. The plugin name doesn't change, from what I can see.

paul-sachs:
@timostamm oh, i definitely misread, sorry. Yeah I can see why moving the rename to a different PR would make sense. I'll update.

paul-sachs:
@timostamm alright, removed package rename (which we'll discuss separately offline).

timostamm (approved):

PR #303, opened 2023-12-13
Added support for cjs output

### LINKED ISSUE #297: Support CommonJS in protoc-gen-connect-query
@bufbuild/protoplugin added support for CommonJS in [v1.6.0](https://github.com/bufbuild/protobuf-es/releases/tag/v1.6.0).

It would be nice if protoc-gen-connect-query supported the this feature as well. It needs to generate export statements with the new method `GeneratedFile.exportDecl` to do so. 

See https://github.com/connectrpc/connect-es/pull/956 for the same update to protoc-gen-connect-es.

paul-sachs:
Fixes #297 

Add support for CommonJS with the new plugin option:

### `js_import_style`

By default, [protoc-gen-connect-query](https://www.npmjs.com/package/@connectrpc/protoc-gen-connect-query)(and all other plugins based on [@bufbuild/protoplugin](https://www.npmjs.com/package/@bufbuild/protoplugin)) generate ECMAScript `import` and `export` statements. For use cases where CommonJS is difficult to avoid, this option can be used to generate CommonJS`require()` calls.

#### Example buf.gen.yaml

```
version: v1
plugins:
  - plugin: es
    out: src/gen
    opt: js_import_style=legacy_commonjs
  - plugin: connect-query
    out: src/gen
    opt: js_import_style=legacy_commonjs
 ```

### CODE REVIEW

File: packages/protoc-gen-connect-query/snapshots/buf.gen.yaml [hunk 1]
```diff
@@ -14,3 +18,10 @@ plugins:
       - target=ts+dts+js
       - import_extension=none
       - ts_nocheck=false
+  - plugin: connect-query
+    path: ../bin/protoc-gen-connect-query
+    out: gen_cjs
+    opt:
+      - target=dts+js
+      - import_extension=.js
+      - js_import_style=legacy_commonjs
```
timostamm, line 27: Shouldn't we set `ts_nocheck=false` here as well, same as we do above?

The snapshots are great, but I find them a bit difficult to follow. Here's the tree:

```
.
├── buf.gen.yaml
├── gen
│   └── connectrpc
│       └── eliza
│           └── v1
│               ├── eliza-ElizaService_connectquery.d.ts
│               ├── eliza-ElizaService_connectquery.js
│               ├── eliza-ElizaService_connectquery.ts
│               └── eliza_pb.ts
└── gen_cjs
    └── connectrpc
        └── eliza
            └── v1
                ├── eliza-ElizaService_connectquery.d.ts
                ├── eliza-ElizaService_connectquery.js
                └── eliza_pb.ts
```

In `gen`, it's a bit odd that we have both ts and d.ts files, and in `gen_js`, we mix js and ts. This may work with our linter, but you would never use this setup in practice, right?

WDYT about making the snapshots more realistic by generating three use cases: ts, js, and cjs?

```yaml
version: v1
plugins:
  # js
  - plugin: es
    out: gen_js
  - plugin: connect-query
    path: ../bin/protoc-gen-connect-query
    out: gen_js

  # cjs
  - plugin: es
    out: gen_cjs
    opt:
      - js_import_style=legacy_commonjs
  - plugin: connect-query
    path: ../bin/protoc-gen-connect-query
    out: gen_cjs
    opt:
      - js_import_style=legacy_commonjs
      - ts_nocheck=false # we only set this for our tests

  # ts
  - plugin: es
    out: gen_ts
    opt:
      - target=ts
  - plugin: connect-query
    path: ../bin/protoc-gen-connect-query
    out: gen_ts
    opt:
      - target=ts
      - ts_nocheck=false # we only set this for our tests
```

timostamm (commented):
Looks great to me, just one suggestion, see below. 

Can we give an example buf.gen.yaml in the next release notes, similar to [this one](https://github.com/connectrpc/connect-es/releases/tag/v1.2.0)?

packages/protoc-gen-connect-query/snapshots/buf.gen.yaml
paul-sachs: Yeah, that does sound like a better option. I'll make the change.

timostamm (approved):

PR #310, opened 2023-12-15
Fix dual package hazard in nodejs

### LINKED ISSUE #305: Getting "No QueryClient set, use QueryClientProvider to set one" error in version 1.0.0
Attempting to use connect-query by following the Quickstart documentation in a Next JS app but having issues. Receiving this error when `useQuery` is called…

```
No QueryClient set, use QueryClientProvider to set one
```

The problem appears to be with the wrapper around the `@tanstack/react-query` `useQuery` as using the original hook does not result in this error. 

My app uses Yarn workspaces so thought it might be an issue with module resolution with my specific environment however I have tried creating a brand new Next JS app with NPM and still getting the same error. I have also tried different node versions and `@tanstack/react-query` versions.

This is the code, which is very close to the example in the docs…

```tsx
import { TransportProvider, useQuery } from "@connectrpc/connect-query";
import { createConnectTransport } from "@connectrpc/connect-web";
import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
import { getUsers } from "clients/dist/user/user-UserService_connectquery";

const finalTransport = createConnectTransport({
  baseUrl: "",
});

const queryClient = new QueryClient();

export default function Web() {
  return (
    <TransportProvider transport={finalTransport}>
      <QueryClientProvider client={queryClient}>
        <Content />
      </QueryClientProvider>
    </TransportProvider>
  );
}

const Content = () => {
  const query = useQuery(getUsers);
  return <div>{query.data?.users.length}</div>;
};
```

```
// package.json
"@bufbuild/protobuf": "~1.3.0",
"@connectrpc/connect-query": "^1.0.0",
"@connectrpc/connect-web": "^1.1.2",
"@tanstack/react-query": "^5.13.4",
```

Any help appreciated! 😄

paul-sachs:
Fixes #305 

For more details, see https://github.com/bufbuild/protobuf-es/issues/610 but the long story short is that module systems in nodejs/browser continue to hurt everyone involved.

### DIFF
--- a/packages/connect-query/package.json
+++ b/packages/connect-query/package.json
@@ -24,9 +24,13 @@
   "main": "./dist/cjs/index.js",
   "exports": {
     ".": {
+      "node": {
+        "import": "./dist/proxy/index.js",
+        "require": "./dist/cjs/index.js"
+      },
       "module": "./dist/esm/index.js",
-      "require": "./dist/cjs/index.js",
-      "import": "./dist/proxy/index.js"
+      "import": "./dist/proxy/index.js",
+      "require": "./dist/cjs/index.js"
     }
   },
   "dependencies": {


timostamm (approved):
Oh, I did not realize we didn't make the update here yet. LGTM!

PR #312, opened 2023-12-19
Remove ESM wrapper

### LINKED ISSUE #305: Getting "No QueryClient set, use QueryClientProvider to set one" error in version 1.0.0

paul-sachs:
Fixes #305... again again.

The `esm` wrapper (as described in https://github.com/bufbuild/protobuf-es/issues/509) creates problems when using a peer dependency that defines it's exports without an import proxy (in our case, react-query). 

# Current situation

Our exports look like:

```json
{
  "exports": {
    ".": {
      "node": {
        "import": "./dist/proxy/index.js",
        "require": "./dist/cjs/index.js"
      },
      "module": "./dist/esm/index.js",
      "import": "./dist/proxy/index.js",
      "require": "./dist/cjs/index.js"
    }
  }
}
```

The above meant the following:
- When the **application** imported `QueryClientProvider` from `@tanstack/react-query`, it resolved to the `esm` version of the context
- When the **library** imported `useQuery` from `@tanstack/react-query`, it resolved instead to the `cjs` version of the package (since NextJS SSR is considered a "node" environment)

This mismatch between the esm and cjs versions causes a problem detecting context

# Alternatives considered

We could expose our own hooks/components with a one-to-one mapping from `@tanstack/react-query` and require our own `QueryClientProvider` to be initialized. This would bypass these kinds of issues, but introduce more setup and potential for confusion. It would mean that using `react-query` for anything non-connect related would require an additional query client provider which MAY actually be the same client depending on where (node/browser) it's running.

# Reason we chose this solution

By not using an `ESM` wrapper, we avoid these possible import mismatches. It does introduce the possibility of dual package hazard with class instances (where instanceof is used, like with ConnectError) but we don't use any of that in connect-query so we can be relatively safe in that regard.


packages/connect-query/package.json
```diff
@@ -22,16 +21,13 @@
...
-        "import": "./dist/proxy/index.js",
-        "require": "./dist/cjs/index.js"
-      },
-      "module": "./dist/esm/index.js",
-      "import": "./dist/proxy/index.js",
+      "import": "./dist/esm/index.js",
       "require": "./dist/cjs/index.js"
-    }
+    },
+    "./package.json": "./package.json"
```
timostamm, line 30: Why export package.json?

packages/connect-query/package.json
```diff
@@ -22,16 +21,13 @@
   "type": "module",
   "sideEffects": false,
   "main": "./dist/cjs/index.js",
+  "module": "./dist/esm/index.js",
```
timostamm, line 24: I think the most simple change would be to delete node from exports, so that we end up with:

```
"exports": {
    ".": {
      "module": "./dist/esm/index.js",
      "import": "./dist/proxy/index.js",
      "require": "./dist/cjs/index.js"
    }
}
```

Note that we're using the module export instead of adding the module field.

It's a good idea to match the upstream package here, but since connect-es and protobuf-es are involved as well, it's better we stick to incremental changes that we can cover in our tests in examples-es.

Note that this change will potentially re-introduce dual package issues around ConnectError. @smaye81 is looking into making the internal usage in ConnectError resilient against dual package issues, which is ultimately the better way forward than the ESM wrappers.

packages/connect-query/package.json
paul-sachs: It's a good callout, I was mostly looking to make things more consistent with react-query, which also exports "./package.json" but I couldn't find the reasoning. I also was reading this [extensive investigation](https://blog.isquaredsoftware.com/2023/08/esm-modernization-lessons/#problems-with-next-js-and-react-server-components) which also had that export (again, with no explanation I could find).

timostamm: You typically export package.json because it's otherwise inaccessible to consumers respecting the exports fields. Unless we have a concrete reason to provide package.json, we should not provide it.

packages/connect-query/package.json
paul-sachs: The additional `module` field was just there to mimic react-query, since it doesn't seem to make a difference with nextjs. The problem is largely that this bug triggers if `module` and `import` don't actually match. So basically, we could do something like:

```ts
"exports": {
    ".": {
      "module": "./dist/esm/index.js",
      "import": "./dist/esm/index.js",
      "require": "./dist/cjs/index.js"
    }
}
```

And it'll work, though now the `module` designation seems redundant.

timostamm: I thought so too, but @smaye81's recent tests in https://github.com/connectrpc/examples-es/pull/1149 proved otherwise.

paul-sachs: Sounds good... well, not good, just confusing, but I'll keep module.

timostamm (approved):

PR #318, opened 2023-12-29
Make sure to take into account enabled queryOption

### LINKED ISSUE #317: `enabled` option does not work as expected
https://github.com/connectrpc/connect-query-es/blob/94ee6e86abbb7fc497c88d5900de457cfc9eb688/packages/connect-query/src/use-query.ts#L61-L64

I think that's the problem.

paul-sachs:
Fixes #317 

Incoming queryOption should be taken into account for final enable status, just don't let it override the disableQuery input.

### DIFF
--- a/packages/connect-query/src/use-query.test.ts
+++ b/packages/connect-query/src/use-query.test.ts
@@ -123,6 +123,42 @@ describe("useQuery", () => {
 
     expect(result.current.data).toBe(6);
   });
+
+  it("can be disabled without explicit disableQuery", () => {
+    const { result } = renderHook(
+      () => {
+        return useQuery(
+          sayMethodDescriptor,
+          {
+            sentence: "hello",
+          },
+          {
+            enabled: false,
+          },
+        );
+      },
+      wrapper({}, mockedElizaTransport),
+    );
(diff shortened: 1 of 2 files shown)


packages/connect-query/src/use-query.test.ts
```diff
@@ -123,6 +123,42 @@ describe("useQuery", () => {
...
+            enabled: false,
+          },
+        );
+      },
+      wrapper({}, mockedElizaTransport),
+    );
+
+    expect(result.current.data).toBeUndefined();
+    expect(result.current.isPending).toBeTruthy();
+    expect(result.current.isFetching).toBeFalsy();
```
smaye81, line 145: Just for my own knowledge, how do we tell that the query was disabled from these values?

paul-sachs: There's currently no explicit indicated of disabled with react-query, but basically if it's `isPending` and `isFetching` is false, then the indication is that it's disabled. If it's enabled, the initial state will be `isPending` is true and `isFetching` = true.

smaye81 (approved):

Issue #328, opened 2024-01-31
Very easy to create collisions between infinite and non-infinite queries


lsmurray:
It is very easy for users to create collisions between infinite and non-infinite queries.

Consider a list API where we use token based pagination. The token is a string and an empty token means we want the first page.
I may call the API with `useQuery` and no `page_token` to get a couple results from the first page.
I may also call the API with `useInfiniteQuery` and a `page_token` for an infinite list of results.
The issue is both these calls produce an identical query key.

You can see why here... the connect code sets `page_token` to `undefined` for infinite query results.

https://github.com/connectrpc/connect-query-es/blob/639d4a529f236a50568021aa4a3db4b0d0dc0105/packages/connect-query/src/create-use-infinite-query-options.ts#L163

Example Proto

```proto
rpc ListUsers(ListUsersRequest) returns (...) {

}

message ListUsersRequest {
  optional string page_token = 1;
}
```

Example keys that are created

```ts
import { hashKey } from "@tanstack/react-query";
const useQueryKey = createConnectQueryKey(listUsers, {
  pageSize: 10,
});

const useInfiniteQueryKey = createConnectQueryKey(listUsers, {
  pageSize: 10,
  pageToken: undefined,
});

console.log(hashKey(useQueryKey) === hashKey(useInfiniteQueryKey))
// true
```

I'm not sure if there is a good solution for this but if there is it should be documented. For now one option is to manually check that every time I call `useQuery` I pass a `page_token` and set it to an empty string.

paul-sachs:
Hmm, this is true. Thanks for bringing this to our attention. I think something that we could consider is having `useInfiniteQuery` use a modified query key, with a trailing parameter to indicate it's for infinite query only. I'll need to investigate the ramifications of that. Since our query keys are all arrays, invalidating should still match without the trailing key, but I'll need to build up some tests to make sure that's the case.

timostamm:
In TanStack Query, query keys are completely up to the user. 

Are there any use cases where you would _want_ to share a cache between an infinite query and a non-infinite query? Or use cases where you would want to use exactly the same RPC, but with _different_ caches?

timostamm:
> Since our query keys are all arrays, invalidating should still match without the trailing key, but I'll need to build up some tests to make sure that's the case.

I'd also expect a query key to invalidate an infinite query that uses the same key with a trailing parameter. Did your tests confirm that, @paul-sachs?

lsmurray:
[...]

PR #339, opened 2024-02-08
Release 1.2.0

paul-sachs:
## What's Changed
* Add a distinguishing factor between infinite query keys and regular query keys by @paul-sachs in https://github.com/connectrpc/connect-query-es/pull/334
* Add support for enabled to useInfiniteQuery by @paul-sachs in https://github.com/connectrpc/connect-query-es/pull/338

## Changes to infinite query keys

With https://github.com/connectrpc/connect-query-es/pull/334, we're fixing a bug where the cache of a regular query collides with the cache of an infinite query, which can lead to type mismatches and unexpected failures if you use the same RPC with an infinite query and a regular query.
We are now using a separate query key for infinite queries to solve the issue (we append "infinite" to the key), but this change can potentially affect you:

If you tried to invalidate some infinite queries like `queryClient.invalidateQueries({ queryKey: createConnectQueryKey(...), exact: true })`, this now only invalidates queries initialized with `useQuery`. To invalidate infinite queries, you can either set `exact` to false, or use `createConnectInfiniteQueryKey()` instead. Also note that this change not only affects `invalidateQueries` but any operations against the queryClient that interact with the cache using a key (eg, queryClient.setQueriesData, etc.)

**Full Changelog**: https://github.com/connectrpc/connect-query-es/compare/v1.1.3...v1.2.0


paul-sachs:
This **may** be worth an update to 2.0 since it does change the way data is stored within react-query cache. Feels like a lot more churn than it's worth however.

timostamm:
> there may be some breakages when using the createConnectQueryKey along with queryClient 

Can you elaborate what this means?

paul-sachs:
@timostamm update description with an example.

timostamm:
Thanks. You mentioned in https://github.com/connectrpc/connect-query-es/issues/328#issuecomment-1932252394:

> having infinite queries share the same key as a regular query leads to very possible type mismatches and unexpected failures

So it's an important bug to fix. The change in behavior you describe is IMO acceptable in order to fix this bug.

I suggest to explain this situation in the release notes, for example:

 > With https://github.com/connectrpc/connect-query-es/pull/334, we're fixing a bug where the cache of a regular query collides with the cache of an infinite query, which can lead to type mismatches and unexpected failures if you use the same RPC with an infinite query _and_ a regular query.
> We are now using a separate query key for infinite queries to solve the issue (we append `"infinite"` to the key), but this change can potentially affect you: If you tried to invalidate...

paul-sachs:
@timostamm awesome suggestion. I've also added a bit making sure it's clear that it's not just `invalidateQueries` that is affected.

PR #343, opened 2024-02-23
Expose internal methods used to create query options

### LINKED ISSUE #296: Availability of createUseQueryOptions and createUseInfiniteQueryOptions in v1
Is the absence of createUseQueryOptions and createUseInfiniteQueryOptions in v1 intentional? These methods appear to be extremely important when React context is not available.

https://github.com/connectrpc/connect-query-es/blob/main/packages/connect-query/src/index.ts#L19-L24

The README mentions createUseQueryOptions:

> What if I have a custom Transport?
> If the Transport attached to React Context via the TransportProvider isn't working for you, then you can override transport at every level. For example, you can pass a custom transport directly to the lowest-level API like useQuery or createUseQueryOptions.
>
> Does this only work with React?
> Connect-Query does require React, but the core (createUseQueryOptions) is not React specific so splitting off a connect-solid-query is possible.

(README Link: https://github.com/connectrpc/connect-query-es?tab=readme-ov-file#does-this-only-work-with-react)

Is there an alternative method now, or was their removal unintentional? If it's an unintentional, I'm willing to help with a PR.

### LINKED ISSUE #333: useQueries with current setup
Hi there!

https://github.com/connectrpc/connect-query-es/issues/67 has previously mentioned using `createQueryOptions` to allow `useQueries` to work, however that was an issue prior to v1 being released.

With the current version of the connect library, what would be the best example to follow to apply `useQueries`?

I imagine `useQueries` types within `@connectrpc/connect-query` would have to match with `@tanstack/react-query`. Is there any other blocker from the architecture perspective that might not allow `useQueries`, for example using more than 1 transport?

paul-sachs:
Ideally we'd be able to create an opinionated version of `useQueries` but due to the complexity of those recursive types and there still being other queryClient methods that can benefit from these APIs, I propose we expose these internal APIs as helpers.

Fixes #333
Fixes #296

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,7 +191,7 @@ function useQuery<I extends Message<I>, O extends Message<O>>(
   options?: {
     transport?: Transport;
     callOptions?: CallOptions;
-  } & UseQueryOptions,
+  } & UseQueryOptions
 ): UseQueryResult<O, ConnectError>;
 ```
 
@@ -219,7 +219,7 @@ function useInfiniteQuery<
     transport?: Transport;
     callOptions?: CallOptions;
     getNextPageParam: GetNextPageParamFunction<PartialMessage<I>[ParamKey], O>;
-  },
+  }
 ): UseInfiniteQueryResult<InfiniteData<O>, ConnectError>;
 ```
 
@@ -252,7 +252,7 @@ Any additional `options` you pass to `useMutation` will be merged with the optio
 ```ts
 function createConnectQueryKey<I extends Message<I>, O extends Message<O>>(
   methodDescriptor: Pick<MethodUnaryDescriptor<I, O>, "I" | "name" | "service">,
-  input?: DisableQuery | PartialMessage<I> | undefined,
+  input?: DisableQuery | PartialMessage<I> | undefined
 ): ConnectQueryKey<I>;
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -313,6 +315,14 @@ queryClient.setQueryData(
 
 ```
 
+### `createUseQueryOptions`
+
+A functional version of the options usually passed the underlying `useQuery` hook. This is useful when interacting with `useQueries` API or queryClient methods (like [ensureQueryData](https://tanstack.com/query/latest/docs/reference/QueryClient#queryclientensurequerydata), etc).
```
timostamm, line 320: The name stutters a bit, and I think it can be confusing that `createUseQueryOptions` is not relevant for `useQuery` from `@connectrpc/connect-query`. What do you think about naming the function `createQueryOptions()`? 

Can we explain that it returns the query options `queryKey`, `queryFn`, and `enabled` for the given RPC? I imagine this would give users a better understanding.

README.md
```diff
@@ -313,6 +315,14 @@ queryClient.setQueryData(
...
+
+### `createUseInfiniteQueryOptions`
+
+A functional version of the options usually passed the underlying `useInfiniteQuery` hook. This is useful when interacting with some queryClient methods (like [ensureQueryData](https://tanstack.com/query/latest/docs/reference/QueryClient#queryclientensurequerydata), etc).
```
timostamm, line 324: Same as above, `createInfiniteQueryOptions`?

README.md
paul-sachs: Yeah, I think I can get behind that. Yeah it's not the most straight forward for sure.

srikrsna-buf (approved):

README.md
timostamm: Thanks! Looks great to me.

timostamm (approved):

PR #351, opened 2024-03-01
Pass abort signal to transport for mutations

paul-sachs:
Need to propagate the abort signal from callOptions into the underlying transport.


timostamm (approved):
LGTM!

PR #369, opened 2024-05-21
Respect enabled:false in QueryClient's default options

anuraaga:
Currently, tanstack's `useQuery` is always passed `false` or `true` for `enabled`. But if a user hasn't provided any explicit enablement, either using `disableQuery` or `queryOptions`, then the `QueryClient`'s default options should be used, which requires making sure `enabled` is unset, not an explicit boolean value.

### DIFF
--- a/packages/connect-query/src/use-query.ts
+++ b/packages/connect-query/src/use-query.ts
@@ -58,14 +58,18 @@ export function useQuery<
     transport: transport ?? transportFromCtx,
     callOptions,
   });
+  const { enabled: baseEnabled, ...baseRest } = baseOptions;
+  const tsOpts = {
+    ...queryOptions,
+    ...baseRest,
+  };
   // The query cannot be enabled if the base options are disabled, regardless of
   // incoming query options.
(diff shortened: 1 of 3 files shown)


packages/connect-query/src/use-query.ts
```diff
@@ -58,14 +58,18 @@ export function useQuery<
...
   // incoming query options.
-  const enabled = baseOptions.enabled && (queryOptions.enabled ?? true);
-  return tsUseQuery({
-    ...queryOptions,
-    ...baseOptions,
-    enabled,
-  });
+  const enabled = baseEnabled ?? queryOptions.enabled;
+  if (enabled !== undefined) {
+    tsOpts.enabled = enabled;
```
anuraaga, line 70: Much of the complexity of the change came from the fact that Typescript doesn't allow assigning `undefined` to a `field?` type of field, actually I had never run into that yet. Let me know if there's a pattern I'm not aware of.

paul-sachs:
@anuraaga Thanks for the PR. I'd never thought to set enabled to false by default but this makes sense. We'll get this merged and released shortly.

Issue #375, opened 2024-06-06
Add API to allow permissive query keys


paul-sachs:
As of version [1.4.1](https://github.com/connectrpc/connect-query-es/releases/tag/v1.4.0), query keys are always generated with their default values populated. While this works really well for preventing duplicate queries, it prevents some use cases of `createConnectQueryKey`, primarily when used along with `queryClient.invalidateQueries`. See [slack thread](https://bufbuild.slack.com/archives/CRZ680FUH/p1717596754080019).

We should expose a standardized way of creating permissive keys that allow more generalized matching where we want to match multiple queries.

errorhandler:
Is there any plans to add this in the short term? We've been holding off updating and I'm evaluating if we should make our own little function for the main use case we have: invalidate all instances of a particular RPC.

timostamm:
Hey Joe, we'll update the function in v2, and we're actively working on v2, so it shouldn't be too long 🙂 

The main reason to cut a v2 is to support [Protobuf-ES v2](https://github.com/bufbuild/protobuf-es/releases/tag/v2.0.0-alpha.1), so we don't plan on changing the connect-query API too much, but this is a good chance to improve `createConnectQueryKey`.

It should be feasible to support creating a key for an entire service, for an individual RPC, and for an RPC with a specific request payload.

BTW, in the interim, you can create a key to invalidate all requests for an RPC with:

```ts
const key = createConnectQueryKey(rpc).slice(0, 2);
```

timostamm:
https://github.com/connectrpc/connect-query-es/pull/416 adds a function to create message keys. It returns plain, stable, JSON serializable objects. If we use it to create query keys, it should solve this issue, and we can remove `defaultOptions.queries.queryKeyHashFn`, and set `queryKeyHashFn: JSON.stringify` per-query as a small optimization.

Can we make other improvements to the query key factories? An overload to invalidate all queries per service?

```ts
declare function createConnectQueryKey(
  method: MethodUnaryDescriptor,
  input?: DisableQuery | MessageInitShape | undefined,
): ConnectQueryKey;
declare function createConnectQueryKey(
  service: MethodUnaryDescriptor,
  input?: DisableQuery | MessageInitShape | undefined,
): ConnectQueryKey;
```

paul-sachs:
[...]

PR #389, opened 2024-07-11
Port to v2

srikrsna-buf:
Port to v2 of connect and protobuf. User code doesn't change unless using some internal types and functions.


itsjoeoui:
Hey @srikrsna-buf , any update on this? Would love to see it merged! 😄 Thanks

timostamm:
@itsjoeoui, we want to release [connect-es](https://github.com/connectrpc/connect-es), connect-query-es, and [connect-playwright-es](https://github.com/connectrpc/connect-playwright-es) with support for v2, and update the [documentation](https://github.com/connectrpc/connectrpc.com) where necessary. We plan to release it at the same time to make it easier to upgrade. That's why this PR isn't active right now, but we are actively working on the release 🙂

itsjoeoui:
> @itsjoeoui, we want to release [connect-es](https://github.com/connectrpc/connect-es), connect-query-es, and [connect-playwright-es](https://github.com/connectrpc/connect-playwright-es) with support for v2, and update the [documentation](https://github.com/connectrpc/connectrpc.com) where necessary. We plan to release it at the same time to make it easier to upgrade. That's why this PR isn't active right now, but we are actively working on the release 🙂

Awesome! Thanks for the update 👍

packages/connect-query/src/connect-query-key.test.ts
```diff
@@ -12,65 +12,49 @@
...
   it("makes a query key with input", () => {
     const key = createConnectQueryKey(methodDescriptor, {
       sentence: "someValue",
     });
     expect(key).toStrictEqual([
-      "service.typeName",
+      ElizaService.typeName,
       "name",
-      { sentence: "someValue" },
+      create(SayRequestSchema, { sentence: "someValue" }),
```
paul-sachs, line 36: This is interesting. I wonder if it's not very useful to store the full `$typeName` in the key. Not that it's a problem really, just feels unnecessary.

packages/connect-query/src/utils.ts
```diff
@@ -59,22 +64,27 @@ export const isAbortController = (input: unknown): input is AbortController => {
...
     updater: ConnectUpdater<O>,
   ) =>
-  (prev?: O): O => {
+  (prev?: MessageShape<O>): MessageShape<O> => {
     if (typeof updater === "function") {
-      return new methodSig.O(updater(prev));
+      return create(
+        methodSig.output,
+        (
+          updater as (prev?: MessageShape<O>) => MessageInitShape<O> | undefined
```
paul-sachs, line 84: Weird cast, is this even needed?

packages/protoc-gen-connect-query/src/generateTs.ts
```diff
@@ -30,33 +27,16 @@ const { safeIdentifier } = codegenInfo;
...
-        f.print("  O: ", method.output, ",");
-        if (method.idempotency !== undefined) {
-          f.print("      idempotency: ", rtMethodIdempotency, ".", MethodIdempotency[method.idempotency], ",");
-        }
-        f.print("  service: {");
-        f.print("    typeName: ", f.string(service.typeName));
-        f.print("  }");
-        f.print("}", isTs ? " as const" : "", ";");
-
+        f.print(f.export("const", safeIdentifier(method.localName)), " = ", f.importSchema(service), ".method.", method.localName, ";");
```
paul-sachs, line 39: This almost makes me think we don't really need our own generator at all, though that would be more of a hassle to migrate.

packages/connect-query/src/utils.ts
timostamm: Yes. 

> Unsafe argument of type `any` assigned to a parameter of type `MessageInitShape<DescMessage & O> | undefined`  @typescript-eslint/no-unsafe-argument

packages/connect-query/src/connect-query-key.test.ts
timostamm: It isn't necessary to store the type name. But it is necessary to handle default values of fields with implicit presence. For example:

```proto
syntax = "proto3";
message M {
  uint32 field = 1;
}
```

The field has an intrinsic default value `0`, and we have to make sure that the following two calls yield the same query key:

```
createConnectQueryKey(foo, {});
createConnectQueryKey(foo, {field: 0});
```

Otherwise, two queries with exactly the same request payload would have different keys.

`create()` sets the property to `0` in both cases, same as the `new` call did previously. Previously, we converted the class to a plain object again, so that it can be hashed deterministically by the [default hash function](https://github.com/TanStack/query/blob/9da00d610ab437ad1da6854c9a5697874ebebb7e/packages/query-core/src/utils.ts#L205), or by stable-hash from our [default options](https://github.com/connectrpc/connect-query-es/blob/6fcd101896b8ba344f174fe9a46a690554e47206/packages/connect-query/src/default-options.ts#L16) (necessary if bigint is used).

In v2, proto3 messages are plain objects. What's missing here is support for proto2 and editions, which will use the prototype chain, and not hash deterministically.

I think we should take care of that separately, with added test coverage. I don't think we need stable-hash and can still play nice with default hash function, and also resolve #375.

paul-sachs: Yeah that definitely makes sense. Leaving the type name works for me.

packages/connect-query/src/utils.ts
paul-sachs: Ah weird. My IDE didn't have any problems. Good to know.

PR #404, opened 2024-09-03
Turn off structural sharing for bigint queries

smaye81:
Tanstack seems to have [added](https://github.com/TanStack/query/pull/7966/files) a noisy `console.error` log for queries that return data that is not JSON-serializable. They recommend to either turn off `structuralSharing` in your queries or to return JSON serializable data from your queryFn.

For the sake of simplicity, we are simply turning off `structuralSharing` in the tests that return a BigInt as part of their response.

### DIFF
--- a/packages/connect-query/src/use-infinite-query.test.ts
+++ b/packages/connect-query/src/use-infinite-query.test.ts
@@ -39,6 +39,16 @@ const methodDescriptor = {
   },
 };
 
+const defaultCallOptions = {
+  // This query returns a ListResponse which contains a bigint.
+  // As a result, tanstack will show an error that
+  // StructuralSharing requires data to be JSON serializable.
+  // So we turn it off for this test. Another option would be to
+  // implement a `select` option to transform the data into something
+  // JSON serializable.
+  structuralSharing: false,
+};
+
 const mockedPaginatedTransport = mockPaginatedTransport();
 
 describe("useInfiniteQuery", () => {
@@ -51,6 +61,7 @@ describe("useInfiniteQuery", () => {
             page: 0n,
           },
           {
+            ...defaultCallOptions,
             getNextPageParam: (lastPage) => lastPage.page + 1n,
             pageParamKey: "page",
           },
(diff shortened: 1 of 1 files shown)


smaye81:
@timostamm was seeing a noisy log after upgrading tanstack-query that said:

```    
StructuralSharing requires data to be JSON serializable. To fix this, turn off structuralSharing or return JSON-serializable data from your queryFn. [@"connectrpc.eliza.v1.PaginatedService","List",#page:0,,"infinite",]: TypeError: Do not know how to serialize a BigInt
```

We could also implement a `select` function here to convert the returned data, so if you'd rather do that, lmk.

paul-sachs:
I think we can enable this in the default options we provide. Just on my phone so don't have a link but it should work if we add it there.

smaye81:
Do you mean adding it [here](https://github.com/connectrpc/connect-query-es/blob/main/packages/connect-query/src/default-options.ts)? 

Would we want to disable this for all queries though by default? Or is this only used on a case-by-case basis and not added automatically to every query.

paul-sachs:
Given the current default options hash function is also only really needed for bigint support, I think it follows that disabling structural sharing would be also be required.

timostamm (changes requested):
Let's take a look at the [feature](https://tanstack.com/query/latest/docs/framework/react/guides/render-optimizations) before we turn it off:

> React Query uses a technique called "structural sharing" to ensure that as many references as possible will be kept intact between re-renders. If data is fetched over the network, usually, you'll get a completely new reference by json parsing the response. However, React Query will keep the original reference if nothing changed in the data. If a subset changed, React Query will keep the unchanged parts and only replace the changed parts.

Given that React Query invalidates queries somewhat regularly, it would be nice not to re-render if the data hasn't changed. This can be implemented with a per-query option that uses protobuf-es' `equals()`.

It might be better to defer until we finish compat with protobuf-es v2, since it will also require changes to query keys and the safe updater. I suggest we create an issue to track support for structural sharing - and mention how to disable the feature.

smaye81:
OK, I'll close this PR and create an issue instead to figure out a better solution.

PR #407, opened 2024-09-09
Remove "module" export condition to prevent confusion with webpack

paul-sachs:


paul-sachs:
I guess we could also move module down below and do something like:

```json
"exports": {
  ".": {
    "import": "./dist/esm/index.js",
    "require": "./dist/cjs/index.cjs",
    "module": "./dist/esm/index.js"
  }
},
```

But i'd rather remove it entirely and not have to worry about it.

timostamm (approved):
CI is busy, but LGTM if it turns green.

PR #416, opened 2024-09-25
V2: Introduce function to create message keys

timostamm:
TanStack Query uses user-provided [query keys](https://tanstack.com/query/latest/docs/framework/react/guides/query-keys) for caching. Keys are hashed into a string, which must be stable (independent of property insertion order).

In v1, we create keys by converting the request message to a plain object via `toPlainMessage`, which has two drawbacks:

1. For refetching queries and similar operations, query keys can also [act as filters](https://tanstack.com/query/latest/docs/framework/react/guides/filters): A filter matches all keys that match the given properties, ignoring other properties. Because `toPlainMessage` includes default values, it's not possible to omit properties. See https://github.com/connectrpc/connect-query-es/issues/375 for details.
2. Because plain messages can include BigInt values which the [default `queryKeyHashFn` implementation](https://github.com/TanStack/query/blob/v5.58.0/packages/query-core/src/utils.ts#L205) does not support, v1 exports a [`defaultOptions` object](https://github.com/connectrpc/connect-query-es/blob/v1.4.2/packages/connect-query/src/default-options.ts#L30) to provide a 3rd party hash function that does support BigInt.

This PR adds a function to convert a Protobuf message to an object that is suitable for a Query Key in TanStack Query:
- Default values are omitted (both implicit and explicit field presence).
- `NaN`, `Infinity`, and `-Infinity` are converted to a string.
- `Uint8Array` is encoded to a string with Base64.
- `BigInt` values are converted to a string.
- Properties are sorted by Protobuf source order.
- Map keys are sorted with `Array.sort`.

Going forward, we can use this function to create query keys, and remove `defaultOptions.queries.queryKeyHashFn`.


packages/connect-query/src/message-key.ts
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * For any given message, create an object that is suitable for a Query Key in
+ * TanStack Query:
+ *
+ * - Default values are omitted (both implicit and explicit field presence).
+ * - NaN, Infinity, and -Infinity are converted to a string.
+ * - Uint8Array is encoded to a string with Base64.
+ * - BigInt values are converted to a string.
+ * - Properties are sorted by Protobuf source order.
+ * - Map keys are sorted with Array.sort.
```
paul-sachs, line 34: Hmm, my assumption with tanstack keys is that they would have handled both of these without requiring explicit sorting being done, but I suppose best to be be sure.

paul-sachs (approved):

packages/connect-query/src/message-key.ts
timostamm: [hashKey does](https://github.com/TanStack/query/blob/9da00d610ab437ad1da6854c9a5697874ebebb7e/packages/query-core/src/utils.ts#L205), yes.

It's no extra cost for us to sort by Protobuf source order. I figure we might just as well sort map keys, and set a per-query `queryKeyHashFn: JSON.stringify`. No point in walking through the object yet another time.

paul-sachs: Yeah makes sense

PR #417, opened 2024-09-25
V2: Rename MethodUnaryDescriptor argument

timostamm:
We're using the names "methodDescriptor", "methodType", and "methodSig" for `MethodUnaryDescriptor`.

For consistency, this changes all names to "schema". We've been using "schema" in protobuf-es, but it's a different use case here. I wouldn't mind an alternative, as long as we're consistent.


paul-sachs (approved):
Yeah this looks good to me

PR #421, opened 2024-09-26
Replace jest with vitest

### LINKED ISSUE #412: Tests for the example emit warnings
`npx turbo run test -F ./examples/react/basic` emits warnings:

```
ts-jest[ts-jest-transformer] (WARN) Got a `.js` file to compile while `allowJs` option is not set to `true` (file: /xyz/packages/connect-query/dist/cjs/index.js). To fix this:
  - if you want TypeScript to process JS files, set `allowJs` to `true` in your TypeScript config (usually tsconfig.json)
  - if you do not want TypeScript to process your `.js` files, in your Jest config change the `transform` key which value is `ts-jest` so that it does not match `.js` files anymore
ts-jest[ts-jest-transformer] (WARN) Got a `.js` file to compile while `allowJs` option is not set to `true` (file: /xyz/packages/connect-query/dist/cjs/connect-query-key.js). To fix this:
  - if you want TypeScript to process JS files, set `allowJs` to `true` in your TypeScript config (usually tsconfig.json)
  - if you do not want TypeScript to process your `.js` files, in your Jest config change the `transform` key which value is `ts-jest` so that it does not match `.js` files anymore
ts-jest[ts-jest-transformer] (WARN) Got a `.js` file to compile while `allowJs` option is not set to `true` (file: /xyz/packages/connect-query/dist/cjs/use-infinite-query.js). To fix this:
  - if you want TypeScript to process JS files, set `allowJs` to `true` in your TypeScript config (usually tsconfig.json)
  - if you do not want TypeScript to process your `.js` files, in your Jest config change the `transform` key which value is `ts-jest` so that it does not match `.js` files anymore
...
```

It would be good to have an example setup that does not emit warnings.

paul-sachs:
Includes all linting changes, helper methods and config files.

This change cleans up a bunch of config and removes some warnings as well. Fixes #412


packages/connect-query/vite.config.ts
```diff
@@ -12,15 +12,20 @@
...
 
-const config: Config = {
-  preset: "../../jest-preset.js",
-  testEnvironment: "@bufbuild/jest-environment-jsdom",
-  moduleNameMapper: {
-    "(.+)\\.js": "$1", // https://connectrpc.com/docs/web/supported-browsers-and-frameworks/#jest
+// https://vitejs.dev/config/
+export default defineConfig({
+  test: {
+    environment: "jsdom",
```
timostamm, line 20: Wow, did jest simply fix their environment? 

https://github.com/vitest-dev/vitest/blob/e6730909b34526eb73b744124f76151624782679/packages/vitest/src/integrations/env/jsdom.ts#L88

timostamm (commented):
Jest is still hugely popular, but I think this is the right choice. Thanks for the migration! I'll merge this into the v2 branch.

timostamm (approved):

packages/connect-query/vite.config.ts
paul-sachs: jest did so much stuff automatically. Eventually, we might take advantage of vitest's browser mode to actually run in the browser (like jasmine... software development is just a circle) and we can stop patching the dom, but for our purposes, this is fine.

timostamm: jest-environment-jsdom provides a DOM so that tests for web apps can run on Node.js. Awesome.

Unfortunately, it also removes the text encoding API from globals, which Node.js has been supporting for many years now. The problem is so bad that it forced us to create `@bufbuild/jest-environment-jsdom`. It's a simple fix, but Jest has been sitting the problem out for years now, blaming jsdom.

Vitest does the same thing with environments, but it fixes the problem (see link above). Kudos to Vitest.

paul-sachs: Ah I understand what you meant now

PR #422, opened 2024-09-26
V2: Remove option for AbortSignal from useMutation

timostamm:
I noticed that `useMutation` accepts `CallOptions.signal`, but other hooks do not. This removes the option. 

I'm not sure why it would be desirable to cancel a mutation. If I'm missing something, please let me know.


paul-sachs (approved):
Makes sense to me.

PR #424, opened 2024-09-26
Remove outdated JSDoc tags

timostamm:
Parts of the API use JSDoc tags, but they are all outdated. That's worse than having no tags, so I propose we remove them.

### DIFF
--- a/packages/connect-query/src/use-infinite-query.ts
+++ b/packages/connect-query/src/use-infinite-query.ts
@@ -35,9 +35,6 @@ import type { DisableQuery } from "./utils.js";
 
 /**
  * Query the method provided. Maps to useInfiniteQuery on tanstack/react-query
- *
- * @param methodSig
- * @returns
  */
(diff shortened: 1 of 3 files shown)


packages/connect-query/src/use-infinite-query.ts
```diff
@@ -35,9 +35,6 @@ import type { DisableQuery } from "./utils.js";
 
 /**
  * Query the method provided. Maps to useInfiniteQuery on tanstack/react-query
- *
- * @param methodSig
- * @returns
```
timostamm, line 40: Quite distracting warnings in JetBrains IDEs:

w

paul-sachs (approved):

packages/connect-query/src/use-infinite-query.ts
paul-sachs: Ick. Yeah let's nuke em

PR #426, opened 2024-09-27
Honor all call options in callUnaryMethod

timostamm:
`callUnaryMethod` accepts `CallOptions` from `@connectrpc/connect`. This updates the implementation to support all options.

### DIFF
--- a/packages/connect-query/src/use-mutation.ts
+++ b/packages/connect-query/src/use-mutation.ts
@@ -21,6 +21,7 @@ import type {
 import { useMutation as tsUseMutation } from "@tanstack/react-query";
 import { useCallback } from "react";
 
+import { callUnaryMethod } from "./call-unary-method.js";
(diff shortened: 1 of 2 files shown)


packages/connect-query/src/use-mutation.ts
```diff
@@ -58,17 +59,11 @@ export function useMutation<
...
-        { typeName: methodSig.service.typeName, methods: {} },
-        methodSig,
-        callOptions?.signal,
-        callOptions?.timeoutMs,
-        callOptions?.headers,
-        input,
-      );
-      return result.message;
-    },
+    async (input: PartialMessage<I>) =>
```
timostamm, line 62: `callUnaryMethod` is used in other hooks. We can use it here as well.

paul-sachs (approved):
Nice!

Issue #418, opened 2024-09-26
Query keys with dynamic transports


Victor-Mestre-Sybo:
Hello, first of all thanks for this amazing project!
We have this setup where our transport contains dynamic parameters, for example:
```
const transport = createConnectTransport({
  baseUrl: `/api/accounts/${accountId}/services`
})
```
So we have this issue where, when switching `accountId` the same cache will be used for a given query since the query key does not contain any information about the transport. We have considered either making our own abstraction of `useQuery` or calling something like `queryClient.removeQueries()` when the transport's baseUrl changes, is there any better alternative (like including this information in the query key)?

timostamm:
Thanks for filing the issue, Victor. `callOptions` are also not part of the query key.

The best workaround is probably to call `createUseQueryOptions`, modify the query key, and pass the options to `useQuery` from `@tanstack/react-query`.

timostamm:
With #437, we can make the Transport part of the query key. It will be important to create the transport in a module scope, or to memoize it. For example:

```ts
const transport = useMemo(() => createConnectTransport({
  baseUrl: `/api/accounts/${accountId}/services`,
}), [accountId]);

useQuery(say, {sentence: "hi"}, {transport});
```

Could be a footgun, but having to keeping track of reference is common in react.

Victor-Mestre-Sybo:
Thank you for your suggestions! 
@timostamm in order to address this issue using the new `createTransportKey` method, I'd still need to call `createUseQueryOptions` and modify the queryKey by prepending the transport key right?

Victor-Mestre-Sybo:
Sorry, I don't really understand it, I guess I'm missing something. How does `createTransportKey` come into play in your proposed solution? 

As far as I understand, by doing `useQuery(say, {sentence: "hi"}, {transport});` , even if a new transport is created, the query key does not depend on the transport, since this is what is used under the hood:
`const queryKey = createConnectQueryKey(methodSig, input);`

Therefore, the cache will be the same for a given key even after changing the transport right?

timostamm:
@Victor-Mestre-Sybo, this PR wires it up: https://github.com/connectrpc/connect-query-es/pull/441

Victor-Mestre-Sybo:
@timostamm now it makes sense! it's looking really promising, thank you so much for addressing this issue!

timostamm:
[...]

PR #427, opened 2024-09-27
V2: Support disabled queries consistently

timostamm:
https://github.com/connectrpc/connect-query-es/pull/369 made a change to `useQuery` to support edge cases with the `enabled` option.

This makes the same change to the other query hooks.

### DIFF
--- a/packages/connect-query/src/create-use-query-options.ts
+++ b/packages/connect-query/src/create-use-query-options.ts
@@ -116,7 +116,7 @@ export function createUseQueryOptions<
   queryKey: ConnectQueryKey<I>;
   queryFn: QueryFunction<MessageShape<O>, ConnectQueryKey<I>>;
   structuralSharing?: Exclude<UseQueryOptions["structuralSharing"], undefined>;
-  enabled: boolean | undefined;
+  enabled?: false;
 } {
   const queryKey = createConnectQueryKey(schema, input);
   const structuralSharing = createStructuralSharing(schema.output);
@@ -127,6 +127,6 @@ export function createUseQueryOptions<
       callOptions,
     }),
     structuralSharing,
-    enabled: input === disableQuery ? false : undefined,
+    ...(input === disableQuery ? { enabled: false } : undefined),
   };
 }
--- a/packages/connect-query/src/use-query.ts
+++ b/packages/connect-query/src/use-query.ts
@@ -59,18 +59,10 @@ export function useQuery<
     transport: transport ?? transportFromCtx,
     callOptions,
   });
-  const { enabled: baseEnabled, ...baseRest } = baseOptions;
-  const tsOpts = {
+  return tsUseQuery({
     ...queryOptions,
-    ...baseRest,
-  };
-  // The query cannot be enabled if the base options are disabled, regardless of
-  // incoming query options.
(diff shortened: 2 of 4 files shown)


packages/connect-query/src/create-use-query-options.ts
```diff
@@ -127,6 +127,6 @@ export function createUseQueryOptions<
       callOptions,
     }),
     structuralSharing,
-    enabled: input === disableQuery ? false : undefined,
+    ...(input === disableQuery ? { enabled: false } : undefined),
```
timostamm, line 130: Changing the logic: Instead of returning `enabled: undefined` from our option factories, we omit the property. 

This means we can use the spread operator to merge our options with user-provided options (see below).

packages/connect-query/src/use-query.ts
```diff
@@ -59,18 +59,10 @@ export function useQuery<
...
-  };
-  // The query cannot be enabled if the base options are disabled, regardless of
-  // incoming query options.
-  const enabled = baseEnabled ?? queryOptions.enabled;
-  if (enabled !== undefined) {
-    tsOpts.enabled = enabled;
-  }
-  return tsUseQuery(tsOpts);
+    ...baseOptions,
+  });
```
timostamm, line 65: Seems easier to read to me.

I wonder if we should change the order here, and let user-provided options override our options. This would provide an escape hatch for things like #418.

TanStack Query let's users provide a `queryHash` even though it's usually computed from the key. Seems fine to me to let users provide a `queryFn`, `queryKey`, or `structuralSharing` in our hooks, even if the typical use case is to compute them.

packages/connect-query/src/create-use-query-options.ts
paul-sachs: The reason this was kept undefined was that enabled can be provided as a default option at the top level (when providing default settings for all queries). This was left as undefined to respect that default. 

Otherwise, yeah spread would've made sense.

timostamm: By returning `undefined`, we do exactly the opposite: We return a property that would override the query client defaults. Then [we undo it](https://github.com/connectrpc/connect-query-es/blob/v1.4.2/packages/connect-query/src/use-query.ts#L69-L71) in `useClient`.

Maybe I'm missing something. Can you elaborate?

paul-sachs: Actually, nvm, this is what I get for reviewing code on my phone. I understand where I got mixed up.

packages/connect-query/src/use-query.ts
paul-sachs: I feel like i'd want a more tailored way to customize generated keys rather than just overriding. If our intention is to build a typesafe version of react-query, then allowing customization of queryFn could be a problem, not to mention what happens if you have different call sites using different versions of that queryFn. Worth thinking through but I do like the cleanup anyways.

paul-sachs (approved):

packages/connect-query/src/use-query.ts
timostamm: Let's discuss in https://github.com/connectrpc/connect-query-es/pull/428

PR #428, opened 2024-09-27
V2: Replace disableQuery with skipToken

timostamm:
TanStack query added [skipToken](https://tanstack.com/query/latest/docs/framework/react/guides/disabling-queries/#typesafe-disabling-of-queries-using-skiptoken) - we can use it instead of `disableQuery`.


packages/connect-query/src/use-query.ts
```diff
@@ -60,8 +60,8 @@ export function useQuery<
     callOptions,
   });
   return tsUseQuery({
-    ...queryOptions,
     ...baseOptions,
+    ...queryOptions,
```
timostamm, line 64: This change accepts all TanStack Query options, even if they override an option provided by us. The types are changed to allow all options.

`queryFn` and `queryKey` are still typed, since the option type is not inferred.

timostamm: @paul-sachs, I've reverted the change to the types in https://github.com/connectrpc/connect-query-es/pull/428/commits/786fde43808967eda6690a4123e7e15a15a5dfa7. 

This means we do _not_ allow overriding computed options with this PR. We can still merge the users options into our options in the hooks because TanStack Query honors the `skipToken` on it's own and we don't have to set `enabled: false`. We can allow overriding later if needed.

paul-sachs: Makes sense to me, might as well keep them in separate PRs anyways.

paul-sachs (approved):

PR #436, opened 2024-10-03
V2: Remove CallOptions

timostamm:
Hooks and several other functions accept [`CallOptions`](https://github.com/connectrpc/connect-es/blob/v1.5.0/packages/connect/src/call-options.ts#L21) from `@connectrpc/connect`. In general, every Connect client should accept a call options argument, but they don't work well with TanStack Query's caching. 

For now, we're removing `CallOptions` from the function arguments in v2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,8 +25,8 @@ Connect-Query is an wrapper around [TanStack Query](https://tanstack.com/query) 
   - [`createConnectQueryKey`](#createconnectquerykey)
   - [`callUnaryMethod`](#callunarymethod)
   - [`createProtobufSafeUpdater`](#createprotobufsafeupdater)
-  - [`createUseQueryOptions`](#createusequeryoptions)
-  - [`createUseInfiniteQueryOptions`](#createuseinfinitequeryoptions)
+  - [`createQueryOptions`](#createqueryoptions)
+  - [`createInfiniteQueryOptions`](#createinfinitequeryoptions)
   - [`ConnectQueryKey`](#connectquerykey)
 
 ## Quickstart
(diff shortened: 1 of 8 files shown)


packages/connect-query/src/call-unary-method.ts
```diff
@@ -18,36 +18,32 @@ import type {
...
 
 /**
  * Call a unary method given its signature and input.
  */
+// eslint-disable-next-line @typescript-eslint/max-params -- 4th param is optional
 export async function callUnaryMethod<
   I extends DescMessage,
   O extends DescMessage,
 >(
+  transport: Transport,
```
timostamm, line 33: Removing `CallOptions` is a breaking change, but we're also changing the function signature here. Transport is strictly necessary to make a call, so I think it's a better fit to let it be a required argument than part of an options argument.

README.md
```diff
@@ -317,60 +316,59 @@ queryClient.setQueryData(
...
-    createQueryOptions(example, { sentence: "Second query" }, { transport }),
+    createUseQueryOptions(example, { sentence: "First query" }, { transport }),
+    createUseQueryOptions(example, { sentence: "Second query" }, { transport }),
   ]);
   ...
 };
 ```
 
-### `createInfiniteQueryOptions`
+### `createUseInfiniteQueryOptions`
```
timostamm, line 356: Oof, I just realized that this function is exported as `createInfiniteQueryOptions`. I'll revert.

What was the plan here, @paul-sachs?

timostamm: I was wondering why the function is exported with a different name. Why not rename the function if this name is a better fit?

paul-sachs: Ah, that I don't remember. No explicit reason. Renaming it internally definitely makes sense.

timostamm: 👍 I'll do that separately.

paul-sachs (approved):

PR #438, opened 2024-10-03
V2: Rename symbols to match exported names

timostamm:
This renames functions and types to match their export names. For context, see https://github.com/connectrpc/connect-query-es/pull/436#discussion_r1786200957.

The first two commits rename symbols. The other commits co-locate the option types with their hooks and simplify them.


paul-sachs (approved):
Nice! Some of this had to be done for creating connect-query-core, so it's nice to get it done here first.

PR #445, opened 2024-10-08
V2: Remove defaultOptions

timostamm:
`@connectrpc/connect-query` exports `defaultOptions` for the QueryClient. In v1, it provides `queryKeyHashFn` with a hash implementation that support BigInt (npm package "stable-hash").

In v2, query keys do not include BigInt values anymore, and the `defaultOptions` object is empty. This removes the export.


paul-sachs (approved):
🥳

PR #447, opened 2024-10-08
V2: Update createProtobufSafeUpdater

### LINKED ISSUE #446: Update createProtobufSafeUpdater for protobuf-es v2
The README gives the following example for `createProtobufSafeUpdater`:

```ts
import { createProtobufSafeUpdater } from '@connectrpc/connect-query';
import { useQueryClient } from "@tanstack/react-query";

...
const queryClient = useQueryClient();
queryClient.setQueryData(
  createConnectQueryKey(example),
  createProtobufSafeUpdater(example, (prev) => {
    return {
      ...prev,
      completed: true,
    };
  })
);

```

The argument `prev` is a generated message shape or undefined, and the return type is a message init shape:

```ts
export type ConnectUpdater<O extends DescMessage> =
  | MessageInitShape<O>
  | ((prev?: MessageShape<O>) => MessageInitShape<O> | undefined);
```

Because the generated message shape contains the property `$typeName` and the optional property `$unknown`, the object returned in the example doesn't satisfy the init shape, and it fails to compile with a complex type error.

timostamm:
Closes #446. The signature of the provided updater function is changed to return a message instead of a message init shape.

### DIFF
--- a/packages/connect-query/src/utils.test.ts
+++ b/packages/connect-query/src/utils.test.ts
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { create, isMessage } from "@bufbuild/protobuf";
+import { create, isFieldSet, isMessage } from "@bufbuild/protobuf";
 import { describe, expect, it } from "vitest";
 
 import { Proto2MessageSchema } from "./gen/proto2_pb.js";
@@ -137,10 +137,14 @@ describe("createProtobufSafeUpdater", () => {
       const next = safeUpdater(prev);
       expect(next).toBeDefined();
     });
-    it("updates message", () => {
+    it("returns message", () => {
       const prev = create(Proto2MessageSchema);
       const next = safeUpdater(prev);
       expect(isMessage(next, Proto2MessageSchema)).toBe(true);
+    });
+    it("updates field", () => {
+      const prev = create(Proto2MessageSchema);
+      const next = safeUpdater(prev);
       expect(next?.int32Field).toBe(999);
     });
     it("keeps existing fields", () => {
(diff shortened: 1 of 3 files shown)


packages/connect-query/src/utils.ts
```diff
@@ -56,25 +56,26 @@ export const isAbortController = (input: unknown): input is AbortController => {
  */
 export type ConnectUpdater<O extends DescMessage> =
   | MessageInitShape<O>
-  | ((prev?: MessageShape<O>) => MessageInitShape<O> | undefined);
+  | undefined
+  | ((prev?: MessageShape<O>) => MessageShape<O> | undefined);
```
timostamm, line 60: Adding `undefined` on line 59 supports a use-case with `QueryClient.setQueryData`: 

> If the value is undefined, the query data is not updated. 

This means the following example compiles and works as expected (query data is not updated):

```ts
  schema: ElizaService.method.say,
  input: { sentence: "hi" },
  transport: myTransport,
});

const newData: SayResponse | undefined = undefined;

const updater = createProtobufSafeUpdater(
  ElizaService.method.say,
  newData,
);

queryClient.setQueryData(queryKey, updater);
```

packages/connect-query/src/utils.test.ts
```diff
@@ -79,57 +75,112 @@ describe("isAbortController", () => {
...
+    const schema = { output: Proto2MessageSchema };
+    const safeUpdater = createProtobufSafeUpdater(schema, (prev) => {
+      if (prev === undefined) {
+        return undefined;
+      }
+      return {
+        ...prev,
+        int32Field: 999,
+      };
+    });
```
timostamm, line 128: We have to explicitly handle the `undefined` case now. In this example, we do not update query data if it does not exist. 

If the goal is to create query data if none exists, the function could be updated to:

```ts
const safeUpdater = createProtobufSafeUpdater(schema, (prev) => {
  prev ??= create(schema.output);
  return {
    ...prev,
    int32Field: 999,
  };
});
```

We could move this logic into `createProtobufSafeUpdater`, and always provide data to the updater function. But then it isn't possible to update only existing query data. I'm not sure what's idiomatic here. WDYT, @paul-sachs?

paul-sachs: I think this new version is more idiomatic to react-query so I'm good with this updated behaviour.

paul-sachs (approved):

PR #448, opened 2024-10-08
V2: Replace `MethodUnaryDescriptor` with `DescMethodUnary` from `@bufbuild/protobuf`

srikrsna-buf:
Replace `MethodUnaryDescriptor` with `DescMethodUnary` from `@bufbuild/protobuf`.


examples/react/basic/package.json
```diff
@@ -26,10 +26,10 @@
     "@testing-library/react": "^16.0.1",
     "@types/react": "^18.3.5",
     "@types/react-dom": "^18.3.0",
-    "@vitejs/plugin-react": "^4.3.1",
+    "@vitejs/plugin-react": "^4.3.2",
```
paul-sachs, line 29: Did we mean to update these dependencies here? Are they somehow related to protobuf updates?

packages/connect-query/src/create-query-options.ts
```diff
@@ -69,7 +69,7 @@ export function createQueryOptions<
   queryKeyHashFn: (queryKey: QueryKey) => string;
 } {
   const queryKey = createConnectQueryKey({
-    schema,
+    schema: schema,
```
paul-sachs, line 72: Unnecessary change.

examples/react/basic/package.json
srikrsna-buf: Ah yes sorry, I forgot to add context. When I tried to update protobuf version, pnpm was downloading types from two different vite versions, which was resulting in a type error in the example's vite config. I tried the usual steps of deleting the lockfile and removing node modules didn't work. So I upgraded to latest.

paul-sachs: Makes sense.

paul-sachs (approved):

timostamm:
Updating the docs in https://github.com/connectrpc/connect-query-es/pull/449/commits/f7c95a2648d8b41b9db430cefceec7e755f1a089.

PR #449, opened 2024-10-08
V2: Update docs

timostamm:


README.md
```diff
@@ -261,27 +267,81 @@ function createConnectQueryKey<Desc extends DescMethod | DescService>(
...
+
+```ts
+import { createConnectQueryKey } from "@connectrpc/connect-query";
+import { ListService } from "./gen/list_pb";
+
+// The hook useInfiniteQuery() creates a query key with cardinality: "infinite",
+// and passes on the pageParamKey.
+const queryKey = createConnectQueryKey({
+  schema: ListService.method.list,
+  cardinality: "infinite", // "any" matches infinite and finite queries
```
paul-sachs, line 333: Not sure what this means by "any". I think if we leave off cardinality, it'll match both. But that would be undefined no?

paul-sachs: Ah I see. Yes I was expecting cardinality for the keys to not default to finite. I actually think default to undefined cardinality is more consistent with existing behaviour since the default was to ignore the final "infinite" section anyways. I can make this change.

paul-sachs (approved):

README.md
paul-sachs: Hmm, unless you were using it for queries (and not for queryClient.invalidateQueries, or other queryClient operations). So there's some subtle breaking changes either way.

timostamm: If we want to support a filter that only matches non-finite queries (related: https://github.com/connectrpc/connect-query-es/issues/328), we have to always include a cardinality property with useQuery and useInfiniteQuery, because omitting a property in a query filter means matching any query.

The current approach seems to be a bit confusing, so we should either:
1. Always include cardinality explicitly, which means users have to remember to pass it for exact query key matching.
2. Revert, and only include cardinality for infinite queries, which means non-infinite queries can not be isolated in filters.

Expanding on 1: Currently, all parameters to `createConnectQueryKey` are optional, which makes it very easy to forget passing `cardinality` (or `transport`, for that matter), which won't match with exact query filters. So it could be an option to provide two versions of the function - one that requires all parameters to create a full key, and one that doesn't ("createPartialQueryKey"?), and is can be used for query filters.

paul-sachs: I prefer 1 for sure, i’m tempted to keep it simple though, without a new API. It feels like just another API that’s so close to the original. An opinionated QueryClient will make the distinction and use cases much more clear and concise.

PR #450, opened 2024-10-09
V2: Enforce cardinality as required prop for key

paul-sachs:
As discussed in [discussion thread](https://github.com/connectrpc/connect-query-es/pull/449#discussion_r1793551960), we need to be explicit about cardinality of query keys.

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,11 +281,11 @@ import { ElizaService } from "./gen/eliza_pb";
 
 const myTransport = useTransport();
 const queryKey = createConnectQueryKey({
-  // The schema is the only required parameter.
   schema: ElizaService.method.say,
   transport: myTransport,
   // You can provide a partial message here.
   input: { sentence: "hello" },
+  // This defines what kind of request it is (either for an infinite or finite query).
   cardinality: "finite",
 });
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -280,24 +281,25 @@ import { ElizaService } from "./gen/eliza_pb";
 
 const myTransport = useTransport();
 const queryKey = createConnectQueryKey({
-  // The schema is the only required parameter. 
+  // The schema is the only required parameter.
```
timostamm, line 284: That's not true anymore.

packages/connect-query/src/connect-query-key.test.ts
```diff
@@ -90,17 +93,11 @@ describe("createConnectQueryKey", () => {
...
     });
     expect(key[1].input).toBe("skipped");
   });
 
-  it("sets cardinality finite by default", () => {
-    const key = createConnectQueryKey({
-      schema: ElizaService.method.say,
-    });
-    expect(key[1].cardinality).toBe("finite");
-  });
```
timostamm, line 102: Maybe it would be helpful to cover that omitting cardinality in the params doesn't set it in the key?

paul-sachs: I think this is actually covered in `allows to set cardinality: undefined`. Although the way that I built the types, you actually can't omit cardinality, it is mandatory (even if it's set to undefined).

timostamm (approved):

PR #452, opened 2024-10-10
V2: Update peer dependencies

paul-sachs:


timostamm (approved):
Good catch. We should cut a release just for this fix.

PR #458, opened 2024-10-14
V2: Fix and update peer dependencies

timostamm:


packages/connect-query/package.json
```diff
@@ -45,7 +45,7 @@
   },
   "peerDependencies": {
     "@bufbuild/protobuf": "2.x",
-    "@connectrpc/connect": "2.x",
+    "@connectrpc/connect": "^2.0.0-beta.2",
```
timostamm, line 48: The constraint `2.x` does not match pre-release versions.

The new constraint `^2.0.0-beta.2` will also match future v2 beta versions. Updating to the second beta of connect-es everywhere.

smaye81 (approved):

PR #461, opened 2024-10-17
Remove query fn provided by createQueryOptions

paul-sachs:


timostamm (approved):
This looks like a shortcoming in TanStack / Query to me, but removing the property is not a big deal, it was just a very small perf improvement.

paul-sachs:
> This looks like a shortcoming in TanStack / Query to me, but removing the property is not a big deal, it was just a very small perf improvement.

This is correct. The APIs in question (getQueryData and setQueryData) don't have a way of providing a custom queryKeyHashFn despite `useQuery` having a way to define it. We may be able to enable this with something like [`setQueryDefaults`](https://tanstack.com/query/latest/docs/reference/QueryClient/#queryclientsetquerydefaults) but for now, this unblocks us.

paul-sachs:
I can confirm that this could be fixed with this:

```tsx
queryClient.setQueryDefaults(["connect-query"], {
  queryKeyHashFn: JSON.stringify,
});
return queryClient.getQueryData(...)
```

Which we could reintroduce if we had a custom query client. It seems that the tanstack team seems hesitate to fix this in a cleaner way: https://github.com/TanStack/query/issues/2352#issuecomment-2100689045

timostamm:
Sure, but this may have side-effects. Not worth it for the very small perf improvement.

PR #466, opened 2024-10-19
refactor: Maintain consistent naming convention for Options across codebase

Lennon57:
While analyzing this library at my new company, I noticed an opportunity for improvement in maintaining consistent naming conventions.

## Changes
- Renamed 'tsOpts' to 'tsOptions' to align with existing naming patterns
- This change ensures consistency with other Options usage

## Why
- Maintains consistent naming convention throughout the codebase
- All other option-related variables already use the full 'Options' suffix
- This change completes the standardization of our options naming pattern

### DIFF
--- a/packages/connect-query/src/use-query.ts
+++ b/packages/connect-query/src/use-query.ts
@@ -56,7 +56,7 @@ export function useQuery<
     callOptions,
   });
   const { enabled: baseEnabled, ...baseRest } = baseOptions;
-  const tsOpts = {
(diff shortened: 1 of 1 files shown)


paul-sachs:
@Lennon57 thank you for the submission but a lot of this code has changed (for the better) in v2 so we're not going to be merging this (to avoid unnecessary merge conflicts). Stay tuned, v2 is very close 🚀

Issue #467, opened 2024-10-23
Return query key from useQuery


anuraaga:
`useQuery` is the primary entrypoint for fetching data and manages creating a query key and setting the query function as per the README. I was wondering if it would make sense to have it return the query key it created? It feels a bit tedious to have to create it again in mutations.

```ts
const queryClient = useQueryClient();

const { data: getStuffRes, queryKey: getStuffKey, isPending } = useQuery(getStuff, {
    stuffId,
  });

// No need to call createConnectQueryKey(getStuff, { stuffId }), which repeats args to useQuery

const doAddStuff = useMutation(addStuff, {
  onSuccess: (resp) => {
    queryClient.setQueryData(getStuffKey, (prev) => {
      return {
        ...prev,
        stuff: [...prev.stuff, resp.newStuff],
     };
    }
  }
});
```

paul-sachs:
Hi @anuraaga, this is an interesting proposal. We could of course do it ourselves, although this feels more like something that would benefit all of @tanstack/query. In fact, this item was already brought up [here](https://github.com/TanStack/query/discussions/6752). Part of me wants to hew close to whatever tanstack query does but I can see the value here.

I think the only argument I really see against return the queryKey is that that key is not disjoint from the associated data stored in the cache. Which means that it will allow you to do something like:

```ts
queryClient.setQueryData(getStuffKey, { somethingTotallyWrong: true });
```

Which will cause everything to break.

We have some potential work coming after the v2 release that eases this slightly with a custom queryClient that would instead allow you to do something like:

```tsx
queryClient.setConnectQueryData(getStuff, { stuffId }, (prev) => {
      return {
        ...prev,
        stuff: [...prev.stuff, resp.newStuff],
     };
    });
```

This method enforces type safety and makes sure that the new data incoming matches the cached data types. It still requires you building the same input, but I personally think that's pretty minor compared to ensuring static type safety.

anuraaga:
Thanks for the consideration @paul-sachs. TBH I have only every used react query with connect, so I'm not sure on idiomatic vanilla usage but IIUC there it didn't make much sense since the user creates the key, and it's trivial to just reference that without returning it. Here, since connect-query is creating it, it could make sense to return what it created.

The final blog post linked in the discussion was interesting

https://tkdodo.eu/blog/the-query-options-api#datatag

[...]

Issue #391, opened 2024-07-12
Add way to provider updater types to queryclient.setQueryData with createConnectQueryKey


nickzelei:
Hey all, I'm playing with the `createConnectQueryKey` and I'm wondering if there is a way to allow TS to better infer the return type when using it in conjunction with the tanstack `queryclient.setQueryData`.

It's able to infer the some of the types, but not all of them. Arguably the most important being the actual value that the cache is going to be updated to.

Screenshot below shows an example. TS is definitely inferring some of the types, but the `updater` is still set to `unknown`. 
It would be awesome if by passing in the query key if the `updater` could be set to the response type, in this case, `GetAccountOnboardingConfigResponse`

<img width="1051" alt="image" src="https://github.com/user-attachments/assets/f8053613-e72c-4038-a189-7969eb840b5f">

timostamm:
I don't think TanStack's `QueryClient.setQueryData` - which we are using here - can infer data type from the key type. To enforce matching key and value, we'd have to provide our own API to set query data.

Issue #390, opened 2024-07-12
Add ability to easily update query cache


nickzelei:
Hey all, awesome project! This is allowing me to easily switch away from `swr` over to using `tanstack/query`.

One thing I miss however is the ability to easily update the query cache.

With swr, each instance has a `mutate` function that includes the curried key, so I can easily pass it the new value.
With the current iteration of this library, the `refetch` function that is returned from `useQuery` takes no parameters.

In order for me to update the underlying cache, I have to do something like this:

```
const queryclient = useQueryClient();
const { mutateAsync } = useMutation(createMyValue);

async function onSubmit(values: MyFormValues): Promise<void> {
  const resp = await mutateAsync(...values...);
  queryclient.setQueryData(
    createConnectQueryKey(geyMyValue, {
      id: resp.myvalue.id,
    }),
    new GetMyValueResponse({
      myvalue: resp.myvalue,
    })
  );
}
```

I'd love it if there were an easier way to trigger a cache update for a `useQuery` value that didn't involve so much boilerplate.

timostamm:
Thanks for the issue! In general, invalidating queries (triggering a refresh) instead of manipulating the cache is preferable because it's so simple and reliable (for example, you'll never forget to also update a `lastUpdated` property).

For example, I've often seen a pattern like this:

```ts
  const res = useMutation(createMyValue, {
    onSuccess: (data) =>
      queryClient.invalidateQueries({
        queryKey: createConnectQueryKey(geyMyValue, {
          id: data.id,
        }),
      })
  });
```

I'm not closely familiar with swr. Is it more common to update the cache than to let it update?

nickzelei:
Thanks for the response! Unclear if it's more common or not to do that. `swr` works effectively the same way as the `refetch` does with their `mutate`, but they optionally allow you to provide a cache value.

Since our mutation APIs like Create, Update return the latest object, I like to optimistically update the cache because we have the latest object instead of just calling refetch again to hit the API one more time.

Perhaps this is an anti-patten and it's just way more common to simply call refetch or invalidate.

timostamm:
> Since our mutation APIs like Create, Update return the latest object, I like to optimistically update the cache

I'm not sure I'd call it an anti-pattern, just that I would generally recommend to invalidate and refetch for simplicity. Makes perfect sense to me to not waste a roundtrip, if you know what you're doing 🙂 

[...]

PR #468, opened 2024-10-24
Introduce custom QueryClient

paul-sachs:
The new QueryClient introduces typesafe APIs to all the relevant queryClient APIs that you'd want to use.

This change can help alleviate the following issues: #467 #390 #391


packages/connect-query-core/src/create-infinite-query-options.ts
```diff
@@ -117,7 +116,6 @@ export function createInfiniteQueryOptions<
     | SkipToken;
   structuralSharing: (oldData: unknown, newData: unknown) => unknown;
   initialPageParam: MessageInitShape<I>[ParamKey];
-  queryKeyHashFn: (queryKey: QueryKey) => string;
```
timostamm, line 120: Can we remove this in a separate PR? We can't take it back after publishing v2.

paul-sachs: Filed [here](https://github.com/connectrpc/connect-query-es/pull/470)

paul-sachs:
@timostamm Moved in suggested changes from #518 into this PR instead. It does feel a lot better to have it integrated into the query client

paul-sachs:
Superseded by #532. Eventually there may be a place for a custom client but for most things, #532 will suffice with much less additional API surface.

PR #485, opened 2024-11-26
Allow omitting cardinality when creating query key

anuraaga:
`cardinality` allows setting `undefined` but requires being set. It seems easier to allow omission for users that don't differentiate between finite and infinite queries in their code

### DIFF
--- a/packages/connect-query-core/src/connect-query-key.ts
+++ b/packages/connect-query-core/src/connect-query-key.ts
@@ -92,7 +92,7 @@ type KeyParams<Desc extends DescMethod | DescService> = Desc extends DescMethod
       /**
        * Set `cardinality` in the key - undefined is used for filters to match both finite and infinite queries.
        */
-      cardinality: "finite" | "infinite" | undefined;
+      cardinality?: "finite" | "infinite" | undefined;
       /**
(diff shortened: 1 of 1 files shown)


paul-sachs:
Hi @anuraaga. Appreciate the time you put into this PR but we had decided to keep this parameter explicit intentionally. The idea was that since the cardinality changes how the information is shaped and sorted for queries, the developer should be explicit. This makes it a little more cumbersome for query client operations but that's why I'm pitching the [custom query client](https://github.com/connectrpc/connect-query-es/pull/468) to make that a little more digestible.

anuraaga:
Ok @paul-sachs. I understand the motivation somewhat but do think that passing `undefined` explicitly is uncommon and doesn't look like idiomatic typescript IMO. If wanting to be explicit probably undefined should have been left out, but that would be a breaking change so not worth it now I guess.

Hopefully the custom client will clean things up.

paul-sachs:
Yeah that is a fair statement. This API has grown to be pretty complicated due to increased complexity in the keys themselves. We should have spent time making sure all parameters were as explicit as possible, with readable defaults. I do believe having an opinionated query client will help to clear up the more complicated internal bits so we appreciate your patience.

PR #493, opened 2024-12-09
Ensure createconnectQueryKey validates input types

paul-sachs:
- Add test to make sure invalid keys are impossible
- Ensure createConnectQueryKey enforces type on input


paul-sachs:
If there's a more typescript-foo way to do this without adding some additional type params, all up for it.

timostamm (approved):
Oops. LGTM.

PR #496, opened 2024-12-11
Match release process with connect-es

paul-sachs:
To better match the release process with connect-es. Left the examples specifically pointing at connect-es tag since we don't really have a good example in connect-query-es.


paul-sachs:
I basically missed this change from https://github.com/connectrpc/connect-query-es/pull/490

timostamm (approved):
TY

PR #497, opened 2024-12-12
Improved workspace version update to support private packages

paul-sachs:
Just pushing some updates after running through a release and noting some rough edges with scripts.


paul-sachs:
Marking as draft until release is done in case there are more changes.

timostamm (commented):
I think there's another issue: The function readPackages expects a directory with packages. It doesn't recognize packages/examples/react/basic as a package.

timostamm:
> I think there's another issue: The function readPackages expects a directory with packages. It doesn't recognize packages/examples/react/basic as a package.

The current Node.js LTS - v22 - implements glob. This makes it easy to read all packages:

```js
import { readFileSync, existsSync, globSync } from "node:fs";
import { dirname, join } from "node:path";

/**
 * Read the given root package.json file, and return an array of workspace
 * packages.
 *
 * @param {string} rootPackageJsonPath
 * @return {{path: string, pkg: Record<string, unknown>}[]}
 */
function findWorkspacePackages(rootPackageJsonPath) {
  const root = JSON.parse(readFileSync(rootPackageJsonPath, "utf-8"));
  if (!Array.isArray(root.workspaces) || root.workspaces.some(w => typeof w !== "string")) {
    throw new Error(`Missing or malformed "workspaces" array in ${rootPackageJsonPath}`);
  }
  const rootDir = dirname(rootPackageJsonPath);
  return root.workspaces
    .flatMap(ws => globSync(join(rootDir, ws, "package.json")))
    .filter(path => existsSync(path))
    .map(path => {
      const pkg = JSON.parse(readFileSync(path, "utf-8"));
      return { path, pkg };
    });
}
```

We can bump to Node v22 (we can still use older versions in CI) and use that. Need to update `writePackages` a bit. I think it's worth it.

paul-sachs:
Closing in preference to #499

paul-sachs:
Reopening since I realized #499 was merging into this branch 😵

PR #498, opened 2024-12-12
Release 2.0.1

paul-sachs:
## What's Changed
* Ensure createconnectQueryKey validates input types by @paul-sachs in https://github.com/connectrpc/connect-query-es/pull/493
* Update dependencies to support react 19 by @paul-sachs in https://github.com/connectrpc/connect-query-es/pull/495

**Full Changelog**: https://github.com/connectrpc/connect-query-es/compare/v2.0.0...v2.0.1


paul-sachs:
Included react 19 specifically since it feels like it's nice to show in release notes and we had active questions about it, even if it was just a dependency update. Not a strong opinion, open to removing it.

scripts/set-workspace-version.js
```diff
@@ -75,10 +75,6 @@ try {
   process.exit(1);
 }
 
-/**
- * @typedef {{path: string; pkg: Package}} Workspace
- */
-
```
timostamm, line 81: Bad merge?

timostamm (commented):
packages/test-utils in package-lock.json wasn't updated. Can you run `node scripts/set-workspace-version.js 2.0.0 && node scripts/set-workspace-version.js 2.0.1 &&` to fix it?

scripts/set-workspace-version.js
paul-sachs: Ah yeah, probably. Oddly, i did run the `npm run setversion` script. I'll try again and see what missed.

timostamm: Yes, but the package was already at 2.0.1, and the lock file wasn't updated. This problem only exists because the old script changed things partially.

paul-sachs: Ah right. That's why we set to 2.0.0, makes sense.

timostamm (approved):

timostamm (approved):

PR #499, opened 2024-12-12
Support npm workspaces in set-workspace-version.js

timostamm:
Support npm workspaces - including patterns in set-workspace-version.js. Update logic for syncing the lock file to handle local packages without a name. 

Context (and base): https://github.com/connectrpc/connect-query-es/pull/497


timostamm:
@paul-sachs, this might work, but let's make sure. Going to give it a test in the other repos tomorrow.

paul-sachs (approved):
Tested locally and it works well for our purposes.

timostamm:
I've refactored the script in https://github.com/connectrpc/connect-query-es/pull/499/commits/16f4364b1aa57209cd75e311acacb57bd40bf642. It requires a package-lock.json file now, and checks conditions more rigorously. Workspace packages can omit a version, but only if they are private.

Also updating the script in https://github.com/connectrpc/connect-es/pull/1368 and https://github.com/bufbuild/protobuf-es/pull/1036.

paul-sachs (approved):

PR #518, opened 2025-02-10
Add utility helper to update infinite data

### LINKED ISSUE #517: createProtobufSafeUpdater and useInfiniteQuery?
types provided from createProtobufSafeUpdater are incompatible with actual data from infinite query

createProtobufSafeUpdater says it will be `QUERY_RESPONSE_TYPE` but in reality the type is `InfiniteData<QUERY_RESPONSE_TYPE>`

```ts
const queryClient = useQueryClient();

const mutation = useMutation(MUTATION_DESCRIPTOR, {
    onSuccess() {
        const updater = createProtobufSafeUpdater(QUERY_DESCRIPTOR, (prevData?: InfiniteData<QUERY_RESPONSE_TYPE>) => {
              return prevData
        });

        queryClient.setQueriesData({ queryKey: QUERY_KEY }, updater);
    },
});
```

Is there a way to provide correct types/arguments?

paul-sachs:
This new API is very similar to `createProtobufSafeUpdater`, just designed to update and validate paginated data.

Fixes #517

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ Connect-Query is an wrapper around [TanStack Query](https://tanstack.com/query) 
   - [`createConnectQueryKey`](#createconnectquerykey)
   - [`callUnaryMethod`](#callunarymethod)
   - [`createProtobufSafeUpdater`](#createprotobufsafeupdater)
+  - [`createProtobufSafeInfiniteUpdater`](#createprotobufsafeinfiniteupdater)
   - [`createQueryOptions`](#createqueryoptions)
   - [`createInfiniteQueryOptions`](#createinfinitequeryoptions)
   - [`addStaticKeyToTransport`](#addstatickeytotransport)
@@ -386,6 +387,10 @@ queryClient.setQueryData(
 );
 
 ```
+
+### `createProtobufSafeInfiniteUpdater`
+
+Creates a typesafe updater for infinite queries. Identical to `createProtobufSafeUpdater` except the data recieved must be of the shape expected for infinite data.
 
 ### `createQueryOptions`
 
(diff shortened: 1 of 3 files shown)


timostamm (commented):
Is it really a good choice to grow this part of the API?

Here's a diff between using createProtobufSafeUpdater, and not using the function:

```diff
queryClient.setQueryData(
  createConnectQueryKey({
    schema: say,
    transport,
    input: {},
    cardinality: "finite",
  }),
- createProtobufSafeUpdater(say, (prev) => {
+ (prev: SayResponse | undefined) => {
    if (prev === undefined) {
      return undefined;
    }
    return create(SayResponseSchema, {
      sentence: prev.sentence + "x",
    });
- }),
+ },
);
```

It's certainly nice that the function enforces the correct return type, but it's still possible to pass in the wrong schema, or set the wrong cardinality for the key.

Wouldn't it be sufficient to document an example for infinite queries, such as:

```ts
queryClient.setQueryData(
  createConnectQueryKey({
    schema: say,
    transport,
    input: {},
    cardinality: "infinite",
  }),
  (prev: InfiniteData<SayResponse> | undefined) => {
    if (prev === undefined) {
      return undefined;
    }
    return {
      pageParams: prev.pageParams,
      pages: prev.pages.concat(
        // data in the cache is always a full message, not an init object
        create(SayResponseSchema, {
          sentence: "x",
        }),
      ),
    };
  },
);
```

paul-sachs:
@timostamm this is fair, the API is a better fit inside a dedicated QueryClient that can enforce data matches. This was just a stepping stone to be able todeprecate both APIs but if we want to focus on getting a dedicated QueryClient out instead, which does a better job of exactly this thing, I can get down with that.

paul-sachs:
Closing in favour of integrating into QueryClient #468

PR #521, opened 2025-02-27
Add support for context values

paul-sachs:
- Add some initial implementation
- Remove useContextValues


paul-sachs:
Closing this for now. Passing context values in a dynamic way has a number of ways that are unpredictable. If you need to set context values, create a custom context and use [addStaticKeyToTransport](https://github.com/connectrpc/connect-query-es?tab=readme-ov-file#addstatickeytotransport) to bind that context value to a query.

PR #532, opened 2025-04-29
Add datatag to ConnectQueryKey so queryClient can infer type info

paul-sachs:
By adding DataTag annotations to the ConnectQueryKey, QueryClient methods can now infer the return type when calling things like `getQueryData` or `invalidateQueries`.

This is at least a partial alternative to #468 which provides better inference abilities while keeping our api surface smaller. There may still be value down the line with very well defined query client methods but I suspect this change covers 90% of the current use cases.

### DIFF
--- a/packages/connect-query-core/src/connect-query-key.ts
+++ b/packages/connect-query-core/src/connect-query-key.ts
@@ -45,35 +45,40 @@ import { createTransportKey } from "./transport-key.js";
  *   }
  * ]
  */
-export type ConnectQueryKey<OutputMessage extends DescMessage> = DataTag<[
-  /**
-   * To distinguish Connect query keys from other query keys, they always start with the string "connect-query".
-   */
-  "connect-query",
-  {
-    /**
-     * A key for a Transport reference, created with createTransportKey().
-     */
-    transport?: string;
-    /**
-     * The name of the service, e.g. connectrpc.eliza.v1.ElizaService
-     */
-    serviceName: string;
-    /**
-     * The name of the method, e.g. Say.
-     */
-    methodName?: string;
-    /**
-     * A key for the request message, created with createMessageKey(),
-     * or "skipped".
-     */
-    input?: Record<string, unknown> | "skipped";
-    /**
-     * Whether this is an infinite query, or a regular one.
-     */
-    cardinality?: "infinite" | "finite" | undefined;
-  },
-], MessageShape<OutputMessage>, ConnectError>;
+export type ConnectQueryKey<OutputMessage extends DescMessage = DescMessage> =
+  DataTag<
+    [
+      /**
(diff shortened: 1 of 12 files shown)


paul-sachs:
We have an open question for this change. We can enforce that all ConnectQueryKeys have this type parameter or just make `createConnectQueryKey` return the tagged type. As it stands, requiring the type tag will be a breaking type change but probably makes sense,

paul-sachs:
I went with an in-between option, I require a param on `ConnectQueryKey` but it defaults to `DescMessage`. This allows people to use `ConnectQueryKey` without the output param but won't be able to infer the return type

packages/connect-query-core/src/connect-query-key.ts
```diff
@@ -18,13 +18,60 @@ import type {
   DescMethodUnary,
   DescService,
   MessageInitShape,
+  MessageShape,
 } from "@bufbuild/protobuf";
-import type { Transport } from "@connectrpc/connect";
-import type { SkipToken } from "@tanstack/query-core";
+import type { ConnectError, Transport } from "@connectrpc/connect";
+import type { DataTag, InfiniteData, SkipToken } from "@tanstack/query-core";
```
timostamm, line 24: `@connectrpc/connect-query` has a peer dependency on `@tanstack/react-query` 5.x. It looks like `DataTag` shipped with [v5.0.0](https://github.com/TanStack/query/releases/tag/v5.0.0). It was merged a week earlier in https://github.com/TanStack/query/pull/6119.

But `TError` on `DataTag` was added later, in https://github.com/TanStack/query/pull/8361. I think it's fair to update the peer dependency constraint. WDYT?

packages/connect-query-core/src/connect-query-key.ts
```diff
@@ -44,35 +91,15 @@ import { createTransportKey } from "./transport-key.js";
...
-     * or "skipped".
-     */
-    input?: Record<string, unknown> | "skipped";
-    /**
-     * Whether this is an infinite query, or a regular one.
-     */
-    cardinality?: "infinite" | "finite" | undefined;
-  },
-];
+export type ConnectQueryKey<OutputMessage extends DescMessage = DescMessage> =
```
timostamm, line 94: We should update https://github.com/connectrpc/connect-query-es/blob/main/README.md#connectquerykey

Not sure we can show the full type anymore, but maybe the literal example we give is sufficient.

timostamm (commented):
Neat!

packages/connect-query-core/src/connect-query-key.ts
paul-sachs: Yep, makes sense to me. I'll update it.

timostamm (approved):
LGTM!

PR #544, opened 2025-07-02
Release workflow

paul-sachs:
This is a modified version of a local script i've been using for a few releases. Hasn't been tested as an action yet so we'll need a new release to validate.

#### Prepare Release Workflow:
* Added `.github/workflows/prepare-release.yml` to automate the creation of draft releases, release branches, and pull requests for new versions. Includes support for potential release/hotfix base branches in the case of a backport fix. To create a backport, we'll just need to create a branch for the target base version (release/v1.x) and set the base branch for the action.

#### Publish Release Workflow:
* Added `.github/workflows/publish-release.yml` to handle the publication of merged release pull requests, including publishing to npm, updating GitHub release notes

#### Update Fixed Issues Workflow:
* Added `.github/workflows/updated-fixed-issues.yml` to comment on fixed issues.


paul-sachs:
After this is merged, I can run it through some dry runs of the first prepare stage which can be considered a noop if the branch is closed with accepting.

paul-sachs:
Alright, updated. After a little bit of click ops on each package, following [these instructions](https://docs.npmjs.com/trusted-publishers#configuring-trusted-publishing), we no longer need a token and we get providence and everything.

paul-sachs:
Confirmed working in this repo: https://github.com/paul-sachs/npm-publish-test. Some minor adjustments have been made to fix cleanup and I've removed the fixed issues action since it wasn't triggering as a cascading action (github prevents cascading actions).

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,87 @@
...
+          cache: "npm"
+
+      - name: Validate semver version
+        run: |
+          if ! echo "${{ inputs.version }}" | grep -E '^[0-9]+\.[0-9]+\.[0-9]+(-[a-zA-Z0-9]+(\.[a-zA-Z0-9]+)*)?(\+[a-zA-Z0-9]+(\.[a-zA-Z0-9]+)*)?$'; then
+            echo "Error: '${{ inputs.version }}' is not a valid semver version"
+            echo "Expected format: MAJOR.MINOR.PATCH (e.g., 1.2.3) or with pre-release/build metadata (e.g., 1.2.3-alpha.1+build.1)"
+            exit 1
+          fi
+          echo "✓ Version '${{ inputs.version }}' is valid semver"
```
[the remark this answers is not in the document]
paul-sachs: Sure. I've updated `set-workspace-version.js` to have an improved regexp and also add an error specifically if the version is invalid. Should be sufficient for now.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,78 @@
...
+          node-version-file: ".nvmrc"
+          cache: "npm"
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Create draft release
+        run: gh release create v${{ inputs.version }} --draft --generate-notes
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
timostamm: Is there a reason to create a draft release? Publishing the draft won't publish, editing the draft doesn't work either, and we have to delete the draft if we abort a release. So I only see downsides to creating a draft.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,78 @@
...
+        run: |
+          npm run setversion ${{ inputs.version }}
+
+      - name: Commit version changes
+        run: |
+          git add .
+          git commit -s -m "Release ${{ inputs.version }}"
+          git push --set-upstream origin "release/prep-release-${{ inputs.version }}"
+
+      - name: Get release notes
```
timostamm: Is this the reason to draft a release? Then I suggest that we create a draft, get the generated release notes, and delete the draft.

.github/workflows/add-to-project.yml
timostamm: Why rename this unrelated file? Can you revert?

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,105 @@
...
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Delete release branch
+        continue-on-error: true
+        run: |
+          echo "Deleting release branch ${{ github.event.pull_request.head.ref }}"
+          git push origin --delete ${{ github.event.pull_request.head.ref }}
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
timostamm: Should this be any different from other PRs that are closed?

I would prefer that we don't do this until we find out that we really need it.

timostamm (commented):
Nice. Let's get this merged here, and apply to other relevant repositories after running it for a while. 

scripts/release.js: Can you update the comment that recommends the procedure for releases? 

.github/RELEASING.md: Can you update this document for the new release process?

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,78 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Version to release (e.g. 1.2.3)"
+        required: true
+        type: string
+      base_branch:
+        description: "Base branch for release (e.g. release/v1.x, hotfix/v1.2.x)"
```
timostamm: Can you explain how this would be used?

.github/workflows/add-to-project.yml
paul-sachs: Huh yeah weird, I'll revert this.

.github/workflows/prepare-release.yml
hudlow: Why not kick the action off by manually creating a draft release and editing the release notes up front? The `prepare-release` action could trigger on that event, pick up the release tag, target branch, and release notes, create the release branch and PR, and delete the draft release.

It wouldn't be very different from what you have, but it would make use of the ["new release" UI](https://github.com/connectrpc/connect-query-es/releases/new) rather than the more esoteric UI to trigger a workflow.

One downside I see is that you could accidentally create a non-draft release in the UI... but this could maybe be handled by permissions or even detected and cleaned up by the `prepare-release` workflow?

scripts/release.js
```diff
@@ -12,20 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
```
hudlow, line 13: Why are we keeping this file?

scripts/utils.js
```diff
@@ -0,0 +1,50 @@
+// Copyright 2021-2023 The Connect Authors
```
hudlow, line 1: Do we need this function to be externalized from the script if we delete `scripts/release.js`?

.github/workflows/prepare-release.yml
paul-sachs: The original idea was to leverage a review process, which is more central to this current version (everything pivots off the PR). Creating a release doesn't really have associated workflows so we'd need to do cleanup. But both would obviously work.

scripts/release.js
paul-sachs: Not sure i understand, `npm run release` runs this script

hudlow: Derp, of course — somehow I was thinking you moved `npm publish` directly to the workflow.

scripts/utils.js
hudlow: Disregard.

scripts/release.js
paul-sachs: Admittedly, it is a little annoying how close those two commands are 🤷

.github/workflows/prepare-release.yml
hudlow: Yeah, totally get why you're using a PR for review, and I wasn't suggesting changing that. I was just suggesting using the draft release creation to kick off the flow which would seem marginally more GitHub-idiomatic to me, even though you still wouldn't be publishing the draft release actually created since it would just be tagged from `main` (or from a branch for a backport).

paul-sachs: Ah i think i now understand what you meant. So instead of triggering via workflow run, we'd trigger on a draft release creation. Yeah, has a little more cleanup but i can see why that might be a little more natural.

paul-sachs: @timostamm do you have thoughts on this? Using a draft release as the trigger feels a little more intuitive vs an arbitrary workflow.

timostamm: Can draft releases trigger a workflow?

paul-sachs: Hmm, not clear from the [docs](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#release). It says that a draft release doesn't trigger a created/edited/deleted, which implies `published` but i don't know how you would trigger a published event on a draft release (i guess when it moves from draft to release). So yeah, that would probably be a blocker to using drafts.

hudlow (approved):

timostamm (approved):

PR #547, opened 2025-07-10
Match structural sharing behavior from @tanstack/query-core

### LINKED ISSUE #546: structuralSharing not working with select in useQuery/useInfiniteQuery
Thanks as always for maintaining this great project!

When using the `select` option to transform query data, structural sharing breaks because transformed data fails the isMessage check and always returns a new reference.

### The Problem
When using select to transform data, createStructuralSharing receives non-Message objects (like plain objects, arrays, or primitives), but it only handles Message types:

```
export function createStructuralSharing(
  schema: DescMessage,
  // eslint-disable-next-line @typescript-eslint/no-redundant-type-constituents -- matching the @tanstack/react-query types
): (oldData: unknown | undefined, newData: unknown) => unknown {
  return function (oldData, newData) {
    if (!isMessage(oldData) || !isMessage(newData)) {
      return newData;  // ❌ Problem: Always returns new reference for non-message data
    }
    if (!equals(schema, oldData, newData)) {
      return newData;
    }
    return oldData;
  };
}
```
https://github.com/connectrpc/connect-query-es/blob/3b38421dc336f0f96b68c8d7c6fd5889fcf56a7e/packages/connect-query-core/src/structural-sharing.ts

### Expected Behavior: 
Even when `isMessage` returns `false` (for transformed data from select), if the values are structurally equal, `createStructuralSharing` should return `oldData` to maintain referential stability and prevent unnecessary re-renders. This would ensure that `@tanstack/react-query` structural sharing optimization works for all data types, not just Message types.

paul-sachs:
If we return a non-message (through select), we should be matching the default behavior for react-query, which is to share references where values are the same. Calling replaceEqualDeep matches that behavior. See query-core code [here](https://github.com/TanStack/query/blob/main/packages/query-core/src/utils.ts#L381).

Fixed #546

### DIFF
--- a/packages/connect-query-core/src/structural-sharing.test.ts
+++ b/packages/connect-query-core/src/structural-sharing.test.ts
@@ -64,5 +64,5 @@ describe("structural sharing", () => {
     const newData = { count: 2 };
     const result = fn(oldData, newData);
     expect(result).toBe(oldData);
-  })
+  });
 });


packages/connect-query-core/src/structural-sharing.test.ts
```diff
@@ -28,34 +28,41 @@ describe("structural sharing", () => {
     const oldData = create(schema, { sentence: "hi" });
     const newData = create(schema, { sentence: "hi" });
     const result = fn(oldData, newData);
-    expect(result).toStrictEqual(oldData);
+    expect(result).toBe(oldData);
```
paul-sachs, line 31: Switching all these to `toBe` so it can check referential integrity instead of deep equality.

timostamm (approved):

PR #558, opened 2025-09-04
Add support for headers

### LINKED ISSUE #555: How to pass headers "on the fly"…?
I'm trying to move from my own wrapping with react-query to this lib.

I need to pass headers _on the fly_ per request. Previously I've done that with the "headers" field of the generated client methods from @connectrpc/connect:

```
   MyClient.getSomeThing(
      { ...params},
      { headers: headers(veryDynamicFieldThatChanges) },
    ); 
```

These I manually wrapped with @tanstack/react-query, but I do see a benefit of using this lib, but I can't see that I can pass my headers through?

I did see https://github.com/connectrpc/connect-query-es/issues/519 but maybe op was fine with the transport based one. I doesn't really work for me as these headers are needed on the lowest level - would have to recreate transport on each call which seems _wrong_?

Is there possibly an option I've missed? Maybe my use case is rare…?

paul-sachs:
Adding headers to the query key and fetching operation.

Fixes #555

### DIFF
--- a/packages/connect-query-core/src/connect-query-key.ts
+++ b/packages/connect-query-core/src/connect-query-key.ts
@@ -46,6 +46,7 @@ type SharedConnectQueryOptions = {
   input?: Record<string, unknown> | "skipped";
   /**
    * Headers to be sent with the request.
+   * Note that invalid HTTP header names will raise a TypeError, and that the Set-Cookie header is not supported.
    */
   headers?: Record<string, string>;
 };
@@ -131,6 +132,7 @@ type KeyParamsForMethod<Desc extends DescMethod> = {
   pageParamKey?: keyof MessageInitShape<Desc["input"]>;
   /**
    * Set `headers` in the key.
+   * Note that invalid HTTP header names will raise a TypeError, and that the Set-Cookie header is not supported.
    */
   headers?: HeadersInit;
 };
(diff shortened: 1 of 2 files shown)


packages/connect-query-core/src/connect-query-key.ts
```diff
@@ -246,5 +255,27 @@ export function createConnectQueryKey<
...
+  ) {
+    props.headers = createHeadersKey(params.headers);
+  }
   return ["connect-query", props] as ConnectQueryKey<O>;
 }
+
+/**
+ * Creates a record of headers from a HeadersInit object.
+ */
+function createHeadersKey(headers: HeadersInit): Record<string, string> {
```
timostamm, line 271: We have to make this a bit more robust. Consider the following declarations:

```ts
const a: HeadersInit = {
  foo: "a, b",
};

const b: HeadersInit = [
  ["foo", "a"],
  ["foo", "b"],
];

const c: HeadersInit = [
  ["foo", "a"],
  ["Foo", "b"],
];
```

They are different formulations for exactly the same headers, but would result in different keys. The most reliable and simple way to handle headers init is to normalize via `new Headers()`.

I suggest adding to the JSDoc:
> Note that invalid HTTP header names will raise a TypeError, and that the `Set-Cookie` header is not supported.

PR #559, opened 2025-09-08
Ignore page param when generating infinite key in createInfiniteQueryOptions

paul-sachs:
Depending on the incoming input, infinite query keys should ignore whatever property informs the page to be returned. This was being done properly when calling `callUnaryMathod` but was not being propagated to `createConnectQueryKey`. This means that calls to `getQueryData` or `setQueryData` would not find the data initialized by `useInfiniteQuery`.

The original test failed to catch this case since 0n (and 0) are default values that evaluate to the same key. Adding a test with non-default page param identified the issue.

### DIFF
--- a/packages/connect-query-core/src/create-infinite-query-options.ts
+++ b/packages/connect-query-core/src/create-infinite-query-options.ts
@@ -83,6 +83,7 @@ export interface ConnectInfiniteQueryOptions<
     MessageInitShape<I>[ParamKey],
     MessageShape<O>
   >;
+  headers?: HeadersInit;
 }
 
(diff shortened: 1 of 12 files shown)


packages/connect-query-core/src/create-infinite-query-options.ts
```diff
@@ -185,6 +185,7 @@ export function createInfiniteQueryOptions<
     schema,
     transport,
     input,
+    pageParamKey,
```
paul-sachs, line 188: Technically, the only change that is actually required. Everything else is optional enhancement to make sure this doesn't happen again

timostamm (approved):

PR #572, opened 2026-03-10
Introduce deep page key support

### LINKED ISSUE #569: add support for nested keys in useInfiniteQuery's pageParamKey
In our backend APIs, the page/offset key is not at the root level. Instead, it is nested inside the query key, e.g.
```json
{
    "query": {
        "search": "",
        "page": 1
        "sort": {
            "key": "name",
            "order": "asc"
        }
    }
}
```

We were using version 2.1.1 before and set `query` as `pageParamKey`. It was working before. but after [these changes ](https://github.com/connectrpc/connect-query-es/pull/559), they stopped working.

An ideal fix will be to add support for nested keys in pageParamKey, e.g., `query.page`

paul-sachs:
Introduces the ability to use page key params that aren't just at the top of the input param. We now support dot separated keys and enforce type safety on their values. 

Fixes #569

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,12 +226,10 @@ Identical to useQuery but mapping to the `useSuspenseQuery` hook from [TanStack 
 function useInfiniteQuery<
   I extends DescMessage,
   O extends DescMessage,
-  ParamKey extends keyof MessageInitShape<I>,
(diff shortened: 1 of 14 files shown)


timostamm (approved):
LGTM!

PR #575, opened 2026-03-21
Fix release prep

paul-sachs:
Bug in release prep, when we removed the base_branch parameter, I forgot to remove the usage.


timostamm:
Superseded by https://github.com/connectrpc/connect-query-es/pull/577

PR #577, opened 2026-05-11
Add GitHub publish workflow with trusted publishing

ajeetdsouza:
Adding a GitHub Actions workflow to publish to npm. This uses [trusted publishing](https://docs.npmjs.com/trusted-publishers) to attach [provenance attestations](https://docs.npmjs.com/trusted-publishers#automatic-provenance-generation) to packages, allowing users to verify their authenticity.


timostamm (approved):

timostamm:
Just noticed: There are some pre-existing workflows related to publishing. Will they conflict?

ajeetdsouza:
They shouldn't, but they are redundant - I thought I deleted those. Sending a followup to remove them.

